PR #43, opened 2014-11-07
Regex support

heyLu:
This is not yet ready to merge, doesn't compile yet. (See the error below.)

I also still need to implement `re-groups` and `re-seq`. Additionally, I'd like to implement IMap for the RegexMatcher so you can get start and end positions (and maybe the groups/named groups) from it.


uvtc:
> It might be a better idea to look at this again once the effect-integration branch is merged.

For reference, it was mentioned back in December 2014 in #104 that, afaict, it was decided that the effects branch would not be merged.

PR #152, opened 2015-01-03
Added filesystem functions to handle FS stuff better

thomasmulvaney:
Previously, `pixie.path` had a basic directory walking function which worked, but had no concept of relative or absolute paths and no way to compare them.

When we loaded tests from the `load-paths` it would look through directories which had the same relative path. This meant that every file was being loaded 3 times. Once for each path. Not a big deal in compiled mode, but this made testing slow in interpreted mode.

The following commit adds `pixie.fs` which aims to give tools for dealing with the filesystem.

We can now deal with the filesystem in a slightly nicer way and run interpreted tests faster.


halgari:
It looks like the translation error is due to something being a string that should be unicode or vice-versa.

thomasmulvaney:
Hi Tim,

I have managed to fix most of these issues, but there are some which look
like they are PyPy ones. os.path.basename and os.path.relpath both seem to
have some issues - relpath seems to have issues because it has keyword
arguments (start=). I have removed those two functions and everything
compiles nicely. I'm just rewriting basename in pixie now and have nearly
finished relpath too.

I should be able to push this off shortly.

PR #154, opened 2015-01-06
Added ex-msg to stdlib

thomasmulvaney:
this allows us to get the message out of an exception. Useful for
testing exceptions.


heyLu:
And thanks again! :)

PR #156, opened 2015-01-07
Range can be turned into a lazy-seq

thomasmulvaney:
We can now go directly from a Range to a lazy sequence. I also switched the field declarations from keywords to symbols.


pixie/stdlib.pxi
```diff
@@ -1565,6 +1565,20 @@ For more information, see http://clojure.org/special_forms#binding-forms"}
         val
         nil))))
 
+;; TODO: place withing Range type when protocols support multi-arity functions
```
heyLu: A nitpick: `s/withing/within/`.

Also, you could place it within Range and use `range->seq` as an inner function. Especially because `-seq` has arity 1.

pixie/stdlib.pxi
```diff
@@ -1565,6 +1565,20 @@ For more information, see http://clojure.org/special_forms#binding-forms"}
...
+        (fn range->seq
+          ([self]
+           (let [start (get-field self :start)]
+             (range->seq self start)))
+          ([self curr-num]
+           (let [stop (get-field self :stop)
+                 step (get-field self :step)]
+             (when (or (and (> step 0) (< curr-num stop))
+                       (and (< step 0) (> curr-num stop))
+                       (and (= step 0)))
```
heyLu: The last `and` is unnecessary.

tests/pixie/tests/test-stdlib.pxi
```diff
@@ -282,4 +282,11 @@
...
      (catch e
-       (t/assert= (ex-msg e) "This is an exception"))))
\ No newline at end of file
+       (t/assert= (ex-msg e) "This is an exception"))))
+
+(t/deftest test-range
+  (t/assert= (-seq (range 10))
+             (-seq (-iterator (range 10))
+             (reduce conj nil (range 10))
+             '(0 1 2 3 4 5 6 7 8 9))))
```
heyLu: `assert=` accepts/compares only 2 arguments, e.g. `(assert= 1 1 2 3 4)` wouldn't throw an exception.

I'm not sure why the additional arguments don't result in an exception. It should, I think.

pixie/stdlib.pxi
thomasmulvaney: Good point

tests/pixie/tests/test-stdlib.pxi
thomasmulvaney: Nice catch! It makes sense that it only takes 2 arguments I will just wrap the three statements in (= ...) and make sure that it returns true.

I will add another pull request which makes assert= throw an error when not given 2 arguments.

heyLu: Ah no, I think the interpreter should check the number of arguments passed, not individual functions.

PR #158, opened 2015-01-12
Pxic

halgari:
Allows files to be compiled to .pxic files. These files are prefered above `.pxi` during the loading process, but the compilation process is not automatic. Because the `.pxic` files are a serialization of the Code objects, this allows for faster loading due to the removal of parsing, compilation, and macro expansion steps. The result is a 10x perf boost on loading pure Pixie code. If macros take even longer to run (like those in ffi-infer) the performance gains may be much higher. 

We should add options to Dust to compile all `.pxi`i dep files to `.pxic`


andrewchambers:
Is it ever the case that a macro depends on the an external file to generate code? I suppose it would be unusual and unadvisable.

halgari:
This is why the creation of .pxic files is not automatic (as in Python). Do do that correctly would require building deps trees. So instead, creation of the files is manual (via calling `./pixie-vm -c`) and should be done as the last step in installation or perhaps as part of dust's deps command

PR #163, opened 2015-01-25
fix crash on syntax errors

gyim:
Some exceptions in pixie.vm.reader were not wrapped, which caused expressions like `#invalid` or `"\123"` to crash the interpreter. This pull request fixes the cases where exceptions are thrown explicitly.

I wrote `return throw_syntax_error_with_data(...)` instead of `throw_syntax_error_with_data(...)` in a few places. Although it does not make much sense, it was needed for rpython to compile the code: the typer assumed that the caller function would continue the execution, and complained about things like reading variables before assignment.

### DIFF
--- a/pixie/vm/reader.py
+++ b/pixie/vm/reader.py
@@ -284,7 +284,7 @@ class LiteralStringReader(ReaderHandler):
             try:
                 v = rdr.read()
             except EOFError:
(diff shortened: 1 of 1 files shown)


heyLu:
Thanks! :)

PR #167, opened 2015-01-30
Vector pop push

thomasmulvaney:


pixie/vm/persistent_vector.py
```diff
@@ -350,11 +350,7 @@ def nth(self, i, not_found=nil):
 
     def pop(self):
         self.ensure_editable()
-        affirm(self._cnt != 0, u"Can't pop and empty vector")
-
-        if self._cnt == 0:
-            self._cnt = 0
-            return self
```
heyLu: @halgari Do you know what this was supposed to do?

heyLu:
Is there a clojure equivalent for `push` or is it just `conj`?

thomasmulvaney:
In clojure there is only `conj`. Interestingly, clojures `IPersistentStack` protocol has `peek` and `pop`, perhaps pixie's `IStack` should have `-peek` too?

arrdem:
Adding `-peek` seems entirely reasonable given that `peek`, `push` and `pop` traditionally define stack types although one could argue that `first` of a stack should have the same effect.

halgari:
for vectors, push = conj, but I suppose it could be different for other data structures, so I'm fine with having both for now.

PR #169, opened 2015-02-03
Added python or pypy to prereqs and changed default to python

gigasquid:
Change the defaults to python rather than pypy so that you have to have one less thing installed to build.

Also, the externals already download pypy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ all: help
 
 EXTERNALS=../externals
 
-PYTHON ?= pypy
+PYTHON ?= python
 PYTHONPATH=$$PYTHONPATH:$(EXTERNALS)/pypy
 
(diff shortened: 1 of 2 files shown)


heyLu:
To clarify about downloading pypy: We need the pypy sources (e.g. `rpython.*`) because the vm is based on pypy. (More precisely, based on rpython, which is the vm infrastructure behind pypy.) The `pypy` executable is not downloaded, we'd have to compile pypy first to get it.

The build works with pypy and (c)python. For some tasks it might make a speed difference. For example, `make run_interactive` is faster when using `pypy` on my machine.

gigasquid:
Hmmm - maybe we should change the makefile to default to pypy if it exists and if not then python?  Or alternatively, we can just say to install pypy as a prereq for building?

PR #174, opened 2015-02-11
Penultimate

gigasquid:
At the presentation of Cincy Functional Programmers, we took a tour of Pixie
and decided to add a _penultimate_ function that returns the second to last element in a collection.

Note that Clojure totally doesn't have this.
:sparkles:

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -838,6 +838,13 @@ If further arguments are passed, invokes the method named by symbol, passing the
       (recur (conj res (first coll)) (next coll))
       (seq res))))
 
+(defn penultimate
+  {:doc "Returns the second to last element of the collection, or nil if none."
+   :signatures [[coll]]
+   :added "0.1"}
+  [coll]
+  (last (butlast coll) ))
+
 (defn complement
   {:doc "Given a function, return a new function which takes the same arguments
          but returns the opposite truth value"}
--- a/tests/pixie/tests/test-stdlib.pxi
+++ b/tests/pixie/tests/test-stdlib.pxi
@@ -90,6 +90,18 @@
     (t/assert= (butlast l) res)
     (t/assert= (butlast r) res)))
 
+(t/deftest test-penultimate
+  (let [v [1 2 3 4 5]
+        l '(1 2 3 4 5)
+        r (range 1 6)]
+    (t/assert= (penultimate nil) nil)
+    (t/assert= (penultimate []) nil)
+    (t/assert= (penultimate (range 0 0)) nil)
+    (t/assert= (penultimate v) 4)
+    (t/assert= (penultimate l) 4)
+    (t/assert= (penultimate r) 4)
+    (t/assert= (penultimate [2]) nil)))
+
 (t/deftest test-empty?
   (t/assert= (empty? []) true)
   (t/assert= (empty? '()) true)


st23am:
:+1:

heyLu:
:fireworks:

Looks like you had fun! :)

lgrapenthin:
As an observer and possible future user of this great project, I wish this would have a name like second-last instead. penultimate is a clever name, but it lacks the naming clarity other core fns have. Here are the usage notes from http://en.wiktionary.org/wiki/penultimate: "While the Latinate penultimate is predominant in written works, the traditional English expressions for this idea were last but one and (less often) second last. Following the 1920s, American use has favored next to last to the point that last but one functions as a Britishism. While last but one continues to be somewhat more popular in Britain, however, next to last, second to last, etc. have been gaining in popularity." - Also notice that as a noun the word is tagged as obsolete, rare, unused or uncommon in every reference on the page - Also see http://public.wsu.edu/~brians/errors/penultimate.html

gigasquid:
I am fine with renaming - it was a fun addition during a group hacking session.  But I can see the obvious reasons for wanting a shorter name

halgari:
yeah, it'd be nice to rename these, I'm up for suggestions.

gigasquid:
Hmm.  Naming is hard.  If we want to make it more general like in https://github.com/pixie-lang/pixie/pull/177, it would start to look like nth but work in reverse.

Maybe `last-nth` or `from-last` ?

arrdem:
my initial reaction to both #174 and #177 was that `nth-last` or simply `nth` support for negative indexing ala Python would be more appropriate. These are some fun names, but not something I'd want to use.

gigasquid:
Agree.  How would you see nth working with negative indexing?  If it is negative it simply starts from the end?  What happens if n is greater than the collection?  Does it wrap?

arrdem:
I would argue that positive indexing with `n > (count col)` is an out of bounds error as one would expect, wrapping is non-obvious and could cause weird issues for users who don't expect it. Besides if that's what you really want modulus is sitting right there. Negative indexing would be `(get col (+ (count col) n))`. So `-1` refers to the last element, `-2` the 2nd to last and soforth. `(get col 0)` remains unambiguous. Again, the way that Python does it.

arrdem:
Oh. Missed a case. If `0 < (+ n (count col))` it's also an out of bounds error.

gigasquid:
I think that would be a nice solution.  It would also still work with people porting Clojure code over to pixie.   Nth would still work the Clojure way, but also the extra pythonesque way.

heyLu:
I also like `nth` with negative indexing. Seems like a natural extension of what `nth` already does.

I'm not sure if we should introduce a "marker protocol" for that or something similar? We could just support it on everything or require it for `nth`, but I'm not sure if we would want that.

heyLu:
I wrote a quick version of `nth` with negative indexing: https://github.com/heyLu/pixie/compare/negative-indexing

Turns out we can just redefine `nth`. (Redefining might be a bit confusing, though, because it's originally defined in [pixie/vm/stdlib.py](https://github.com/pixie-lang/pixie/blob/master/pixie/vm/stdlib.py#L289-L291). I'm not sure why, we might be able to just delete the original definition.)

Feel free to play with this. :)

lgrapenthin:
I don't know, but I can hardly imagine writing an algorithm where I'd purposefully make the call to nth with a positive number and later with a negative number in the same position. If this an idiom in Python, I'd be glad if you could point me to one. My worry is that it makes it easier to write an erroneous algorithm  without noticing.

Rather, I'd prefer nth to throw, like in Clojure. My preference for this would be nth-last.

gigasquid:
I am fine with either approach, (nth-last or nth with negative indexing).  I was looking in the Clojure group and negative indexing has come up before in this thread 

https://groups.google.com/forum/#!searchin/clojure/negative$20index$20nth/clojure/3_7O6L9wDes/RevtcsuY7OIJ

Some of the reasons against negative nth in Clojure were:

```
I agree that negative indexing (and presumably also modulo indexing for the upper index?) is very useful. Stuff like this comes up all the time in core.matrix

However I don't think it makes sense as a standard feature in Clojure's low-level data constructs for several reasons:
a) It's a breaking change for people who are handling out-of-bounds cases
b) In new code it will mask out-of-bounds errors, which can conceal some nasty bugs
c) Cramming more implicit features into existing constructs is a bad idea in general: explicit is better
d) It probably adds a small performance overhead. Not worth making everyone pay a cost for one special case feature
```
We don't have to worry about breaking changes or the modulo indexing, but wanted to bring up if the items in b, c, d would apply to pixie?

heyLu:
I think (c) is the most relevant, so I think `nth-last` and `nth-mod` would be good names.

arrdem:
I would argue that (d) is a non-issue, and (a) both isn't possible and the proposed behavior is _very_ well defined.

However I'm with @heyLu, `nth-last` would be a perfectly acceptable name. `nth-mod` implies that it's modulo arithmetic nth, which is very different from the `nth` I sketched at above not that it couldn't also be added.

halgari:
The performance bits won't really matter for pixie. The check for a
negative number will result in an integer guard (which is cheap, that would
be removed if the number could be considered a constant, so (nth coll -1)
would have no performance impact.


Perhaps the best route is to make a new function called "index" or
something that expresses a more feature-rich nth. I like the idea of Python
indexing, and since the performance doesn't matter, this just seems to be a
question of least surprise and or how error prone it would make client
code.


On Thu, Feb 12, 2015 at 8:36 AM, Reid D McKenzie <notifications@github.com>
wrote:

> I would argue that (d) is a non-issue, and (a) both isn't possible and the
> proposed behavior is *very* well defined.
>
> However I'm with @heyLu <https://github.com/heyLu>, nth-last would be a
> perfectly acceptable name. nth-mod implies that it's modulo arithmetic
> nth, which is very different from the nth I sketched at above not that it
> couldn't also be added.
>
> —
> <https://github.com/pixie-lang/pixie/pull/174#issuecomment-74091007>.
>



-- 
programs.”
(Robert Firth)

gigasquid:
That sounds good.  What is the best way to proceed?  Although this pull request sparked the discussion, it feels like it has turned into a new feature.  Should we open a new issue?  I am would be happy to work on it, but if someone wants to tackle it before I carve out some time to get to it, I would not mind at all.

halgari:
sure, open a new issue and reference this thread.

On Thu, Feb 12, 2015 at 11:15 AM, Carin Meier <notifications@github.com>
wrote:

> That sounds good. What is the best way to proceed? Although this pull
> request sparked the discussion, it feels like it has turned into a new
> feature. Should we open a new issue? I am would be happy to work on it, but
> if someone wants to tackle it before I carve out some time to get to it, I
> would not mind at all.
>
> —
> <https://github.com/pixie-lang/pixie/pull/174#issuecomment-74121898>.
>



-- 
programs.”
(Robert Firth)

Issue #175, opened 2015-02-11
Segmentation faults when using "reduce"


yellowbean:
Hi, I assume this is not a bug ? or pixie didn't implement same as clojure does ?
```lisp
(reduce + (range 100))
```
There will be a Segmentation faults error

Regards

heyLu:
I think this is a bug, it should work. I don't know when I have time to look into this, but for now `(reduce + 0 (range 100))` and `(transduce + (range 100))` both work.

I'm not sure why you get a "Segmentation fault" error though, I get an exception complaining about the number of arguments.

yellowbean:
just give some background setting
64bit, Ubunt 14.04 LTS

andrewchambers:
can confirm, segfault in repl with jit enabled on arch linux 64 bit, fresh pypy download.

andrewchambers:
no jit, works fine with runtime exception.

PR #176, opened 2015-02-11
Two arity version of reduce

thomasmulvaney:
Lets you do `(reduce + (range 10))` as mentioned in #175


pixie/stdlib.pxi
```diff
@@ -166,8 +166,11 @@
                           (recur)))))))))
 
 
-(def reduce (fn [rf init col]
-              (-reduce col rf init)))
+(def reduce (fn 
+              ([rf col]
+               (reduce rf (first col) col))
```
halgari: I think this should be `(reduce rf (first col) (next col))`, Otherwise this happens: `(= (reduce + [1 2]) 4)`

heyLu:
An alternative would be the following:

```clojure
(def reduce
  (fn
    ([rf col]
      (reduce rf (rf) col))
    ([rf init col]
      (-reduce col rf init))))
```

This would be a break from Clojure's behaviour, which [we have talked about before](https://botbot.me/freenode/pixie-lang/search/?q=rich), I'm not sure if we still want that.

Any opinions?

thomasmulvaney:
Yeah i was thinking that could be nice too. But what about functions like
conj which don't have a meaningful identity element? The advantage of your
approach is a better error could be returned, for example "Function has no
0 arity version."

halgari:
And that's the crux of the problem. It's not really useful, except for functions like + and *. And for those functions, why not just go with the (rf) approach? So yeah, I'm in favor of heyLu's version.

thomasmulvaney:
Yeah, the approach given by @heyLu seems best to me too. So I will close this pull request. 

To deal with the suggested approach should we add an `-arity?` method to `IFn` so we can do something along the lines of:

```clojure
(if (arity? rf 0) 
   (rf) 
   (throw "Reducing function must support 0 args"))
```

halgari:
It'd be better to just not have it segfault, I'll dig into it and write a PR to throw an exception on the wrong number of args.

PR #177, opened 2015-02-11
Add antepenultimate and preantepenultimate

rab:
While there are several selectors for the initial elements of a collection `[first second third fourth ... ]` and with `last` being joined by the recent inclusion of @gigasquid `penultimate` from the cincyfp meeting, the complements of `third` and `fourth` are now offered as well.

    [ ... preantepenultimate antepenultimate penultimate last]

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -839,11 +839,29 @@ If further arguments are passed, invokes the method named by symbol, passing the
       (seq res))))
 
 (defn penultimate
-  {:doc "Returns the second to last element of the collection, or nil if none."
+  {:doc "Returns the penultimate element from the collection, or nil if fewer than two."
    :signatures [[coll]]
    :added "0.1"}
   [coll]
-  (last (butlast coll) ))
+  (last (butlast coll)))
+
+(defn antepenultimate
+  {:doc "Returns the antepenultimate element from the collection, or nil if fewer than three."
+   :signatures [[coll]]
+   :added "0.1"}
+  [coll]
+  (cond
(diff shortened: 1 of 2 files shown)


lgrapenthin:
I'd recommend to implement this via drop-last if it exists in pixie, for better performance. Btw from a users point of view I'd prefer to type words like second-last, third-last etc. instead of long rarely used undashed latin. This feels a lot like car, cdr, cadr, ....

rab:
I think that you're looking for `butlast` (which is like the inverse of `next`). This is what is used for the implementation of `penultimate`, but it gets complicated for the fact that `(nil? (butlast [1]))` which makes it impossible to directly do something like `(last (butlast (butlast coll)))` when the number of elements of `coll` is fewer than needed.

lgrapenthin:
Nope. butlast is eager and consumes the sequence every time you call it. Thus nesting it in the way proposed makes the fns waste of computing time. It shouldn't be too difficult to write an efficient transducer version of drop-last, which takes a numeric argument, and reusing that instead.

gigasquid:
Talking about a more generalized / modern name over here https://github.com/pixie-lang/pixie/pull/174  - ideas?

gigasquid:
@rab After some discussion, it seems like the best way to go is for a more generalized function `index` that is a more feature rich `nth` than Clojure's.   It is opened up as a new issue.   Would you be interested in tackling it with me?

https://github.com/pixie-lang/pixie/issues/179

halgari:
Closing this, development has moved to issue #179

PR #183, opened 2015-02-13
add while, take-while, drop-while

zachcp:
and some simple tests for them

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1499,11 +1499,13 @@ The new value is thus `(apply f current-value-of-atom args)`."
            ([result] (rf result))
            ([result input]
               (let [drop? @dv]
-                (if (and drop? (pred input))
-                  result
-                  (do
-                    (reset! dv nil)
-                    (rf result input)))))))))
+                (if drop? 
+                  (if (pred input) 
+                    result 
+                    (do 
+                      (reset! dv nil) 
+                      (rf result input))) 
+                  (rf result input))))))))
   ([pred coll]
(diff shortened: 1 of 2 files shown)


pixie/stdlib.pxi
```diff
@@ -1455,6 +1455,63 @@ The new value is thus `(apply f current-value-of-atom args)`."
...
+    `(loop []
+      (when ~test
+        ~@body
+        (recur))))
+
+(defn take-while
+  {:doc "Returns a lazy sequence of successive items from coll while
+        (pred item) returns true. pred must be free of side-effects.
+        Returns a transducer when no collection is provided."
+  :added "10.1"}
```
halgari: 0.1 perhaps?

gigasquid:
Nice.  My only comment is that since the code is handling transducers, we should have some tests for them.

zachcp:
@gigasquid sure. just not really sure how to do that.....  call with no arity and see if it returns a functions.

something like this maybe?:
(t/assert=    (type (take-while pos?)) pixie.stdlib.Closure)

heyLu:
You can test them by using `transduce` directly, using something like this: `(transduce (take-while even?) conj [2 4 6 7 8])`.

zachcp:
thanks @heyLu  I added the transduce examples to the tests.

zachcp:
added more tests to check trandsuce when there is an initial value.

zachcp:
updated the drop-while to avoid multiple lookups/atom resets once the drop condition has been reached. idea/code courtesy of  @lgrapenthin

PR #184, opened 2015-02-13
Fix Gilardi scenario for -e

lxsli:
This didn't work, because -e would only evaluate the first form
provided:

    px -e '(ns c (require pixie.string :as s))
           (println (s/capitalize "yay"))'

Now -e works more similarly to script files, so this works fine.


heyLu:
Thanks! :sparkles:

PR #186, opened 2015-02-17
[Proposed] nth act like clojure's nth

gigasquid:
Currently in pixie - if `nth` is called with a bad index, it returns nil.  It also doesn't handle default values.
The thought is, to make it easier for people to transition Clojure code to pixie, to make `nth` behave more like Clojure version.  Then we can have another function named `index` or `ith` to have a more functionality (like handling negative indexes)

This pull request does the following for `nth`:
* throws an exception "Index of out Range" for a bad index
* does not throw an exception for a bad index if a not-found value is specified, it will return that instead.

Example:

```clojure
user => (nth [1 2 3] 1)
2

user => (nth [1 2 3] 5)
Error:  in <unknown> at 15:1
(nth [1 2 3] 5)
^
in pixie function nth

in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 632:15
  ([coll idx] (-nth coll idx))
              ^
in polymorphic function -nth dispatching on pixie.stdlib.PersistentVector

in internal function _nth

RuntimeException: Index out of Range

user => (nth [1 2 3] 5 :x)
:x
```

** Note1: this is my first time in the python's vm, so any feedback on better implementation is appreciated.
** Note2: If this is not the desired behavior of Pixie, that is totally cool, trying to implement the change was a good learning experience in itself.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -624,6 +624,14 @@ returns true"
   [n]
   (= (rem n 2) 1))
 
+(defn nth
+  {:doc "Returns the element at the idx.  If the index is not found it will return an error.
+         However, if you specify a not-found parameter, it will substitue that instead"
+   :signatures [[coll idx] [coll idx not-found]]
+   :added "0.1"}
+  ([coll idx] (-nth coll idx))
+  ([coll idx not-found] (-nth-not-found coll idx not-found)))
+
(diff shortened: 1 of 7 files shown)


halgari:
I think this change is good. Or at least I can't think of a better alternative at the moment.

PR #190, opened 2015-02-23
implement pixie.string/blank?

kgann:
This adds an implementation of [clojure.string/blank?](https://clojuredocs.org/clojure.string/blank_q). I have no python experience so let me know if the code is not up to par.

I noticed `pixie.stdlib/empty?` is using a question mark while `pixie.string/ends-with` and `pixie.string/starts-with` are not. Is there a desired naming convention for boolean functions in pixie?

### DIFF
--- a/pixie/fs.pxi
+++ b/pixie/fs.pxi
@@ -106,7 +106,7 @@
     (last (string/split (abs this) ".")))
 
   (extension? [this ext]
-    (string/ends-with (abs this) ext))
+    (string/ends-with? (abs this) ext))
 
   IObject
   (-hash [this]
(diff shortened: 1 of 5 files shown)


halgari:
Yes, we should switch `ends-with` to `ends-with?`. And the same with `starts-with?`. 

There's no reason why we can't implement these functions in pure pixie code. Keeping code out of RPython reduces compilation times and makes debugging easier, since you don't have to re-compile as often. Use nth and count to access characters in the list and character literals to check if a given character is whitespace:

Example:
`(= \space (nth " " 0))`

kgann:
I renamed `ends-with` and `starts-with` in a second commit.

Thanks for the info.

PR #192, opened 2015-02-24
Exception printing and querying

gigasquid:
Some basic functions for dealing with exceptions.

This adds a way to print and turn stack traces into seqs.

It adds two functions `pst` and `trace`

`pst` works on either an exception or if no arg is given the last exception thrown as `*e`.  It will simply print the stack trace.

```clojure
user => (/ 0 0)

; orig stack trace omitted

user => (pst)
in <unknown> at 1:1
(/ 0 0)
^
in pixie function 

in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10
  ([x y] (-div x y))
         ^
in internal function _div

RuntimeException: Divide by zero

nil

user => (pst *e)
in <unknown> at 1:1
(/ 0 0)
^
in pixie function 

in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10
  ([x y] (-div x y))
         ^
in internal function _div

RuntimeException: Divide by zero

nil
```

`trace` also works on either an exception or if none is given, the last bound to `*e`.  It turns the trace into a seq, which someone can do further processing and inspection on.

```clojure

user => (trace)
("Divide by zero" "in internal function _div\n" "in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10\n  ([x y] (-div x y))\n         ^" "in pixie function \n" "in <unknown> at 1:1\n(/ 0 0)\n^")

user => (trace *e)
("Divide by zero" "in internal function _div\n" "in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10\n  ([x y] (-div x y))\n         ^" "in pixie function \n" "in <unknown> at 1:1\n(/ 0 0)\n^")

```

`seq` was also extended to Runtime Exceptions.

@halgari Let me know if this is what you were thinking of..

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2114,7 +2114,7 @@ Expands to calls to `extend-type`."
    :signatures [[] [e]]
    :added "0.1"}
   ([] (pst *e))
-  ([e] (when e (ex-pr e))))
+  ([e] (when e (print (str e)))))
 
 (defn trace
   {:doc "Returns a seq of the trace of a Runtime Exception or the last Runtime Exception in *e"
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2114,7 +2114,7 @@ Expands to calls to `extend-type`."
    :signatures [[] [e]]
    :added "0.1"}
   ([] (pst *e))
-  ([e] (when *e (ex-pr *e))))
+  ([e] (when e (ex-pr e))))
 
 (defn trace
   {:doc "Returns a seq of the trace of a Runtime Exception or the last Runtime Exception in *e"
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2100,7 +2100,6 @@ Expands to calls to `extend-type`."
   ([sb] (str sb))
   ([sb item] (conj! sb item)))
 
-
 (defmacro using [bindings & body]
   (let [pairs (partition 2 bindings)
         names (map first pairs)]
@@ -2110,3 +2109,16 @@ Expands to calls to `extend-type`."
                 `(-dispose! ~nm))
               names)
        result#)))
+(defn pst
+  {:doc "Prints the trace of a Runtime Exception if given, or the last Runtime Exception in *e"
+   :signatures [[] [e]]
+   :added "0.1"}
+  ([] (pst *e))
+  ([e] (when *e (ex-pr *e))))
+
+(defn trace
+  {:doc "Returns a seq of the trace of a Runtime Exception or the last Runtime Exception in *e"
+   :signatures [[] [e]]
+   :added "0.1"}
+  ([] (trace *e))
+  ([e] (seq e)))
--- a/pixie/vm/stdlib.py
+++ b/pixie/vm/stdlib.py
@@ -781,15 +781,15 @@ def _seq(self):
     from pixie.vm.keyword import keyword
     assert isinstance(self, RuntimeException)
     trace = vector.EMPTY
-    trace_element = hmap.EMPTY
-    trace_element = rt.assoc(trace_element, keyword(u"type"), rt.wrap(u"runtime"))
+    trace_element = rt.hashmap(keyword(u"type"), keyword(u"runtime"))
     trace_element = rt.assoc(trace_element, keyword(u"data"), rt.wrap(self._data))
     trace = rt.conj(trace, trace_element)
     for x in self._trace:
-        trace_element = hmap.EMPTY
         tmap = x.trace_map()
+        trace_element = hmap.EMPTY
         for key in tmap:
-            trace_element = rt.assoc(trace_element, keyword(key), rt.wrap(tmap[key]))
+            val = tmap[key]
+            trace_element = rt.assoc(trace_element, key, val)
 
         trace = rt.conj(trace, trace_element)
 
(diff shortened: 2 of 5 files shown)


pixie/stdlib.pxi
```diff
@@ -2110,3 +2109,16 @@ Expands to calls to `extend-type`."
                 `(-dispose! ~nm))
               names)
        result#)))
+(defn pst
+  {:doc "Prints the trace of a Runtime Exception if given, or the last Runtime Exception in *e"
+   :signatures [[] [e]]
+   :added "0.1"}
+  ([] (pst *e))
+  ([e] (when *e (ex-pr *e))))
```
kgann: should this be the var `e` instead of `*e`?

gigasquid: yes indeed - nice catch :)

halgari:
BTW, this build failed due to BitBucket being down. Not your fault.

pixie/vm/stdlib.py
```diff
@@ -774,8 +774,23 @@ def _str(self):
...
     return rt.wrap(self.__repr__())
 
+@extend(_seq, RuntimeException)
+def _seq(self):
+    import pixie.vm.persistent_vector as vector
+    assert isinstance(self, RuntimeException)
+    trace = vector.EMPTY
+    trace = rt.conj(trace, self._data)
+    for x in self._trace:
+        trace = rt.conj(trace, rt.wrap(x.__repr__()))
```
halgari: I like this, but I'd also like a bit more detail. Can we add a method to each of the ErrorInfo types (found here: https://github.com/pixie-lang/pixie/blob/master/pixie/vm/object.py#L157), that construct a map or something for each frame? As it stands we'd need to parse the strings if we want anything useful out of the trace. So you could probably just clone each `__repr__` method and construct a map instead of a string. So polymorphic code info would look something like `{:type :polymorphic :name "-count" :type <type int>}`

halgari: BTW, keep the patches coming, it's great to see this stuff moving forward.

gigasquid: :+1:  :sparkles:

gigasquid:
Here is my next pass of exception trace maps:

```clojure
Pixie 0.1 - Interactive REPL
(darwin_x86_64, clang)
:exit-repl or Ctrl-D to quit
----------------------------
user => (/ 0 0)
Error:  in <unknown> at 1:1
(/ 0 0)
^
in pixie function 

in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10
  ([x y] (-div x y))
         ^
in internal function _div

RuntimeException: Divide by zero

user => (trace)
({:data "Divide by zero", :type :runtime} {:type "native", :name "_div"} {:line_number "511", :line "  ([x y] (-div x y))", :column_number "10", :type "interpreter", :file "/Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi"} {:type "pixie", :name ""} {:line_number "1", :line "(/ 0 0)", :column_number "1", :type "interpreter", :file "<unknown>"})
user => 
```

feedback welcome as always

gigasquid:
oh actually saw a minor tweak now that I am looking at the pull request, making the type map values all strings.  fix on the way....

heyLu:
That looks pretty good! :sparkles:

Why did you make `:line_number` and `:column_number` strings? I'd imagine it would be more convenient to get numbers back.

(And even more nitpicky: Could you change the names to `:line-number` and `:column-number` or even `:line` and `:column`? Dashes seem to be the usual Clojure style for keywords...)

gigasquid:
Thanks - great feedback.  I  will make the changes.  I am still struggling working with python/rpython, so I lose track of the little stuff sometimes while just trying to get it to compile and build :)

pixie/stdlib.pxi
```diff
@@ -2110,3 +2109,16 @@ Expands to calls to `extend-type`."
                 `(-dispose! ~nm))
               names)
        result#)))
+(defn pst
+  {:doc "Prints the trace of a Runtime Exception if given, or the last Runtime Exception in *e"
+   :signatures [[] [e]]
+   :added "0.1"}
+  ([] (pst *e))
+  ([e] (when e (ex-pr e))))
```
heyLu: You can already do this with `(print (str e))`, then you don't need a separate `ex-pr` or alternatively implement it as just `(print (str e))`.

(Sorry that this is so late.)

gigasquid: Cool I will update in the other in the other pull request

gigasquid:
Cool - implemented the changes and learned quite a bit about rpython in the process :)

Here is the latest:

* changed the line and column numbers to ints
* changed the types to keywords
* changed pst to use `(pr (str e))`

```clojure
user => (/ 0 0)
Error:  in <unknown> at 1:1
(/ 0 0)
^
in pixie function 

in /Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi at 511:10
  ([x y] (-div x y))
         ^
in internal function _div

RuntimeException: Divide by zero

user => (trace)
({:data "Divide by zero", :type :runtime} {:type :native, :name "_div"} {:line-number 511, :line "  ([x y] (-div x y))", :column-number 10, :type :interpreter, :file "/Users/cmeier/workspace/misc/pixie/pixie/stdlib.pxi"} {:type :pixie, :name ""} {:line-number 1, :line "(/ 0 0)", :column-number 1, :type :interpreter, :file "<unknown>"})
```

heyLu:
Thanks a lot! :sparkles:

Issue #180, opened 2015-02-12
IO Enhancement Request: readlines /line-seq


zachcp:
readlines/line-seq would be great additions to pixie.

I started penciling one out but run into issues: its all a bit new. Instead of using strings is there a way to recognize the byte as end of line?  

```[clojure]
(defn readline
  ([filestream]
    readline filestream [])
  ([filestream acc]
    (let [lineend (hash-set "\n" "\r")
          acc acc
          c (str (-read filestream))]
      (if (contains? lineend c)
       (apply string acc)
       (readline filestream (conj acc c))))))
```

halgari:
You can use `char` to convert ints to characters, `nth` on a string to get a character from it, and `int` will convert a character to a integer. These might help.

kgann:
@zachcp This is implemented in #193

zachcp:
@kgann  nice!

PR #193, opened 2015-02-24
implement pixie.io/line-seq and pixie.io/read-line

kgann:
This adds `read-line` to the `IInputStream` protocol and implements `line-seq`

https://github.com/pixie-lang/pixie/issues/180

### DIFF
--- a/pixie/io.pxi
+++ b/pixie/io.pxi
@@ -12,7 +12,6 @@
 
 (defprotocol IInputStream
   (read-byte [this] "Read a single character")
-  (read-line [this] "Read a single line")
   (read [this buffer len] "Reads multiple bytes into a buffer, returns the number of bytes read"))
 
 (defprotocol IOutputStream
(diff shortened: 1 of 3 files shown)


heyLu:
Nice!

One nitpick/question: Why is `read-line` a new method on `IInputStream`, couldn't it be just a function that takes an `IInputStream` and returns the line?

kgann:
You're correct, it could just be a function. Would you prefer this?

heyLu:
Yes, because when someone writes a `TCPInputStream` or similar that implements `IInputStream` we can reuse `read-line`. :)

kgann:
Great point. See 1849c62b2271a608f07020f10dfde1a023782275

thomasmulvaney:
You could create a `IRead` protocol which extends `IInputStream`

So you could have:
```
(defprotocol IRead
   (read-line [])
   (read-string [])
   (read-sexp []))

(extend-protocol IRead
   IInputStream
   (read-line []
       ;; Implementation... )
   
   String
   (read-line [] ...)
   ...)
```
This might make sense if people want to `read-line` from things which don't implement `IInputStream`.

PR #194, opened 2015-02-25
Return nil from print, println, pr and prn

heyLu:
The previous return value (how many characters were printed) was an implementation detail and possibly a bit confusing. Returning nil is also what clojure does, although very likely for different reasons.


gigasquid:
nice :) I was just wanting that the other day

heyLu:
You're welcome! :)

PR #198, opened 2015-02-26
Fixes namespacing of keywords in reader

thomasmulvaney:


halgari:
Looks great! We'll merge it once the builds finish.

thomasmulvaney:
Trying to get the `::name` style shortcuts in now, but looks like it is a bit more involved.

halgari:
Merging, feel free to open a new PR for the rest.

gigasquid:
@thomasmulvaney nice!  Are you actively working on the `::name` shortcuts?  If not, I could take a crack at it?

thomasmulvaney:
I am but I am at work so you're welcome take a crack! 

A few pointers, you'll want to add a flag on to a keyword in the `reader.py` to say that it needs to be resolved to a namespace. Then `compiler.py` you'll want check if the keyword has the shortcut and resolve it to the namespace of the context.

At least thats what i think would need to be done. Have fun :sparkles:

halgari:
The namespace resolution stuff should probably happen all inside the
reader. There's a dynamic var in the compiler that stores the current NS.
It's bound during compilation, so if you pull that in via rt, you can do
something like rt.ns.deref() inside the reader. Reference how load-paths is
pulled in in rt and you should be able to do it the same way.

On Fri, Feb 27, 2015 at 7:09 AM, Thomas Mulvaney <notifications@github.com>
wrote:

> I am but I am at work so you're welcome take a crack!
>
> A few pointers, you'll want to add a flag on to a keyword in the reader.py
> to say that it needs to be resolved to a namespace. Then compiler.py
> you'll want check if the keyword has the shortcut and resolve it to the
> namespace of the context.
>
> At least thats what i think would need to be done. Have fun [image:
> :sparkles:]
>
> —
> <https://github.com/pixie-lang/pixie/pull/198#issuecomment-76399440>.
>



-- 
programs.”
(Robert Firth)

thomasmulvaney:
Awesome, I had no idea that the reader could have access to that stuff.

gigasquid:
Cool I will take a look over the weekend

Issue #196, opened 2015-02-26
No default identity hash


halgari:
We don't have a default hash for objects. That means that `(hash (var foo))` currently fails. We should default to some sort of hash on the rpython object (`rpython.rlib.objectmodel.compute_identity_hash` might be usable for this).

halgari:
Fixed by PR #199

PR #199, opened 2015-02-28
added an identity hash method to Object

thomasmulvaney:
I'm not sure if this is what you were thinking in #196 but heres an attempt. It wasn't possible to make `pixie.object.Object` extend `IObject` because it has no type and I wasn't sure if that was the best way to approach the problem, so I have just added a check to see if something satisfies `IObject` otherwise `obj_hash` is called.

### DIFF
--- a/pixie/vm/stdlib.py
+++ b/pixie/vm/stdlib.py
@@ -105,8 +105,14 @@ def default_str(x):
     assert isinstance(tp, Type)
     return rt.wrap(u"<inst " + tp._name + u">")
 
+def default_hash(x):
+    tp = x.type()
+    assert isinstance(tp, Type)
+    return x.hash()
+
 _str.set_default_fn(wrap_fn(default_str))
 _repr.set_default_fn(wrap_fn(default_str))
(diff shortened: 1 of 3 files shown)


halgari:
A slightly cleaner route is to use the default_fn behavior of the protocol functions: 

https://github.com/pixie-lang/pixie/blob/master/pixie/vm/stdlib.py#L25

I haven't benchmarked the satisfies? yet so I don't know how well it behaves, but the JIT optimizes default fns quite well.

thomasmulvaney:
Ok great, I've changed it to use `default_fn` option. Looks much nicer.

pixie/vm/stdlib.py
```diff
@@ -105,8 +105,14 @@ def default_str(x):
     assert isinstance(tp, Type)
     return rt.wrap(u"<inst " + tp._name + u">")
 
+def default_hash(x):
+    tp = x.type()
```
halgari: Do we need this? It's going to get removed by the JIT (or even RPython) but it is extra code.

halgari:
thanks!

Issue #195, opened 2015-02-26
Namespaced keywords broken


halgari:
Neither `:foo/bar` or `::foo` work from the repl. We should fix this at some point.

PR #201, opened 2015-02-28
Keyword with current namespace shortcut ::

gigasquid:
```clojure
Pixie 0.1 - Interactive REPL
(darwin_x86_64, clang)
:exit-repl or Ctrl-D to quit
----------------------------
user => :foo
:foo
user => ::foo
:user/foo
```

Feedback is always welcome


halgari:
finishes off #195

halgari:
Thanks!

PR #204, opened 2015-03-01
name is not a method of namespace

thomasmulvaney:


heyLu:
Mhh.. that test failure looks suspiciously like one I had a few days ago. Do the tests pass locally for you?

gigasquid:
Thanks -  I had played around with adding a name method to namespace and forgot to change it back after refactoring.

halgari:
error in build is from a race condition. merging

PR #212, opened 2015-03-05
Fn arg check

### LINKED ISSUE #211: Add arg count checking for fn calls
It might be nice to add arg count checking when too many args are provided in a function call.

e.g. the following:
```
user => (defn sq [x] (* x x))
<inst pixie.stdlib.Var>
user => (sq 4 5)
16
```
should result in an exception of some kind.

Note that providing too few args to a function *is* checked:
```
user => (sq)
Error:  in <unknown> at 11:1
(sq)
^
in pixie function sq

RuntimeException: Invalid number of arguments 0 for function 'sq'. Expected 1

```

thomasmulvaney:
Fixes #211 

I wrote tests and some better checks a while ago, but it seems those tests were rather broken. See `tests-fns.pxi`. Something like `(is (thrown? ...))` from clojure.test would be nice.

This also ended up catching some things with extraneous args.

### DIFF
--- a/tests/pixie/tests/test-stdlib.pxi
+++ b/tests/pixie/tests/test-stdlib.pxi
@@ -311,9 +311,9 @@
   (t/assert= (merge-with identity {} {:a 1, :b 2}) {:a 1, :b 2})
   (t/assert= (merge-with identity {:a 1} {:b 2}) {:a 1, :b 2})
 
-  (t/assert= (merge-with #(identity %1) {:a 1} {:a 2}) {:a 1})
-  (t/assert= (merge-with #(identity %1) {:a 1} {:a 2} {:a 3}) {:a 1})
-  (t/assert= (merge-with #(identity %2) {:a 1} {:a 2}) {:a 2})
+  (t/assert= (merge-with (fn [a b] a) {:a 1} {:a 2}) {:a 1})
+  (t/assert= (merge-with (fn [a b] a) {:a 1} {:a 2} {:a 3}) {:a 1})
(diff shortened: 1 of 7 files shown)


tests/pixie/tests/test-stdlib.pxi
```diff
@@ -333,9 +333,8 @@
 
 (t/deftest test-range
   (t/assert= (= (-seq (range 10))
-                (-seq (-iterator (range 10))
```
thomasmulvaney: Whoops, a missing `)`!

tests/pixie/tests/test-stdlib.pxi
```diff
@@ -333,9 +333,8 @@
...
-                (reduce conj nil (range 10))
```
thomasmulvaney: Having fixed the missing bracket, this test fails because reducing on a list build a list in reverse `(9 8 7 ...)`. So I've removed it. My bad.

pmonks:
> Something like (is (thrown? ...)) from clojure.test would be nice.

+1

Also, I was thinking test-fns.pxi should probably have some tests around multi-arity and variable-arity fns i.e. that these argument count validations don't get triggered.  wdyt?

thomasmulvaney:
I've added in some MultiArityFn and VariadicFn runtime exceptions. The messages are probably not the best atm and I had to touch a fair few files. I'm still not that happy with the tests either, but its something for now.

thomasmulvaney:
@pmonks the false positives you mention are exactly what was wrong with my original tests. I've added an `assert-throws?` macro to `pixie.tests` to make this stuff work properly.

PR #220, opened 2015-03-11
deftype performance boost for immutable fields #215

### LINKED ISSUE #215: Optimize for immutable datastructures
We should optimize get on CustomTypes to use `@elidable_promote` for fields that are never modified. this should be done via a quasi-immutable field. Done correctly the jit should be able to optimize for immutable fields, without the need for hinting by the user.

halgari:
fixes #215


gigasquid:
:+1:

PR #222, opened 2015-03-12
implement doc string support for native functions

halgari:
With this PR, any function that uses `@as_var` or `@wrap_fn` will have its python doc string pulled in as a pixie doc string. Let's start documenting native functions!


arrdem:
:+1:

PR #226, opened 2015-03-14
Allow terminating macro characters as character literals

justinj:
Previously
```
(def a \))
```
would fail to parse. This commit fixes that by making sure the LiteralCharacterReader always reads at least one character.

I saw in another PR you said you've sort of abandoned these tests - is there a more appropriate place to add tests for something like this?


justinj:
It's not clear to me why the Travis build failed, but it doesn't look related to my change (and it passes locally). Is it possible that test is susceptible to random failures?

halgari:
Agreed, I need to fix that race condition.

Issue #219, opened 2015-03-11
Externals should be made within the project


arsenerei:
Pixie shouldn't leak its magic into my other directories.

andrewchambers:
fair enough, tbh I really dislike how it seemed to download some pre compiled libraries on my system even though I can install them myself.

halgari:
If someone wants to update the build scripts so that they work properly on all supported platforms, and fulfill these goals, I'll happily accept the PR.

heyLu:
Fixed in #229

PR #229, opened 2015-03-17
Build externals within project #219

arsenerei:
For #219


halgari:
Surprised it was this simple, I tried this once and thought I had many more issues. Thanks for the work!

PR #233, opened 2015-03-17
Add more comments tests and fix case I missed

justinj:
I fixed the comment problem for lists, vectors, and maps, but missed sets, so I've fixed that now, sorry about that.

I also added a couple tests.


halgari:
There's a merge conflict here, I think it's due to me adding read_inner so that we can add wrap logic inside read. Mind updating it?

justinj:
Fixed, thanks for letting me know.

PR #240, opened 2015-03-20
Adds an evironment associative structure

thomasmulvaney:
The environment stuff is namespaced to `pixie.stdlib` perhaps I should put it under something else though?

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2340,3 +2340,13 @@ Expands to calls to `extend-type`."
 (def *e)
 (defn -set-*e [e]
   (def *e e))
+
+(extend -str Environment
+        (fn [v]
+          (let [entry->str (map (fn [e] (vector (-repr (key e)) " " (-repr (val e)))))]
+            (apply str "#Environment{" (conj (transduce (comp entry->str (interpose [", "]) cat) conj v) "}")))))
+
+(extend -repr Environment
+        (fn [v]
+          (let [entry->str (map (fn [e] (vector (-repr (key e)) " " (-repr (val e)))))]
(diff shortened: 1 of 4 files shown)


heyLu:
I think you could do this entirely in pixie by now, using the `getenv` and `setenv` (or `putenv`) from libc.

thomasmulvaney:
You could get and set via `getenv` and `setenv`, but are you able to get a list of all env vars using libc? I also thought some of the functionality might be useful from with in the rpython code.

pmonks:
Looks like this guy might have it: https://bitbucket.org/pypy/pypy/src/default/rpython/rtyper/module/ll_os_environ.py

heyLu:
Ah right, there *is* `man 7 environ`, but I don't know if our FFI can handle it yet. (It's a `**char`. We'd also have to do the key/value splitting ourselves.)

thomasmulvaney:
@pmonks the call to `os.environ.list()` just wraps around the code you linked to. `os.environ` is actually just a constant in rpython and the dict lookups i'm making are really translated to the calls in that link.

heyLu:
@thomasmulvaney I don't know, maybe ask tbaldridge about it on IRC? (I tried using `pixie.ffi-infer/defconst`, but it only seems to work for integers, not `**char`.)

halgari:
#241 should help here. And yes, it would be nice to keep this in pixie code.

halgari:
This branch can't merge cleanly with master. If someone could update it, I'll merge it.

thomasmulvaney:
I've rebased it. Seems to be the usual failure on Travis. Env should probably be made a global mutable object perhaps and using the mentioned FFI stuff would be ideal too but haven't had a chance to dig into that much.

PR #243, opened 2015-03-21
CSP anyone?

halgari:
PROBLEM: We need a good clean way to communicate between stacklets 

SOLUTION: Implement CSP in the style of core.async. 

This PR includes channels, buffers, alts!, take, put and go blocks. Since we have true continuations, we can do all sorts of fun stuff with CSP. 

Alts are currently not randomly dispatched. We should fix this soon.

### DIFF
--- a/pixie/buffers.pxi
+++ b/pixie/buffers.pxi
@@ -123,6 +123,18 @@
 (defn sliding-buffer [size]
   (->SlidingBuffer (ring-buffer size)))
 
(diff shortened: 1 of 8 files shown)


andrewchambers:
Cool :)

heyLu:
Oh wow, this looks amazing! :fireworks:

thomasmulvaney:
awesome!

PR #245, opened 2015-03-21
Make the REPL history vars work

truecolour:
Currently at some point `*1`, `*2`, `*3` and `*e` are all defined to `nil` (Where is the code that does that?), and stay `nil` when evaluating stuff in the REPL. This PR makes them work like they were presumably intended to.

[Random remark: It looks like you cannot `def` a var in a different namespace than the current one. This is probably a feature :).]

### DIFF
--- a/pixie/repl.pxi
+++ b/pixie/repl.pxi
@@ -21,7 +21,9 @@
       (try (let [form (read rdr false)]
              (if (= form eof)
                (exit 0)
-               (println (eval form))))
+               (let [x (eval form)]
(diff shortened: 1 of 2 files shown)


halgari:
thanks!

heyLu:
The code is for intializing those vars is in [target.py](https://github.com/pixie-lang/pixie/blob/master/target.py#L43-L50). The REPL code used to be there, but it was ported to the separate namespace recently. You *could* probably move the defs into the repl namespace, but there might be some scoping issues then.

PR #253, opened 2015-03-25
Various additions to the standard libraries

truecolour:
This PR adds various constants, functions and macros that I've found useful while doing some toy projects in pixie, and slightly tweaks the behaviour of various others.

I tried to keep the commits mostly independent, so it should be easy to remove stuff if so desired.

I don't know much about PyPy; someone should probably take a closer look at the RPython parts (665e78b and 654829b) before merging them.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -801,7 +801,7 @@ If further arguments are passed, invokes the method named by symbol, passing the
 (defn string? [v] (instance? String v))
 (defn keyword? [v] (instance? Keyword v))
 
-(defn list? [v] (instance? PersistentList v))
+(defn list? [v] (instance? [PersistentList Cons] v))
 (defn set? [v] (instance? PersistentHashSet v))
 (defn map? [v] (satisfies? IMap v))
 (defn fn? [v] (satisfies? IFn v))
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1046,15 +1046,6 @@ Creates new maps if the keys are not present."
                             (assoc m k (apply update-inner-f (get m k) f ks))))]
     (apply update-inner-f m f ks)))
 
-; FIXME: This causes compile_basics to fail on stacklets.pxi for some weird reason.
-; (defn map-invert
-;   {:doc "Return a map where the vals are mapped to the keys."
-;    :examples [["(map-invert {:a :b, :c :d})" nil {:b :a, :d :c}]]}
-;   [m]
-;   (reduce (fn [m* ent]
-;             (assoc m* (val ent) (key ent)))
-;           {} m))
-
 (def subs pixie.string.internal/substring)
 
 (defmacro assert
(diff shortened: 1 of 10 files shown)


pixie/stdlib.pxi
```diff
@@ -1568,8 +1589,21 @@ not enough elements were present."
...
-       (cons (take n s) (partition n step (drop step s))))))
+   (when-let [s (seq coll)]
+     (lazy-seq
+       (cons (take n s) (partition n step (drop step s)))))))
+
+(defn partitionf [f coll]
+  {:doc "A generalized version of partition. Instead of taking a constant number of elements,
+         this function calls f with the remaining collection to determine how many elements to
+         take."
+   :examples [["(fpartition first [1 :a, 2 :a b, 3 :a :b :c])"
```
halgari: `fpartition` vs `partitionf` but I like where you're going with this

pixie/vm/stdlib.py
```diff
@@ -337,16 +337,36 @@ def apply__args(args):
 @returns(bool)
 @as_var("instance?")
 def _instance(c, o):
-    affirm(isinstance(c, Type), u"c must be a type")
+    from pixie.vm.persistent_vector import PersistentVector
+    if isinstance(c, Type):
+        return true if istypeinstance(o, c) else false
+    elif isinstance(c, PersistentVector):
```
halgari: I'd prefer this be a check that something satisfies both ICounted and IIndexed. That way it supports things like arrays as well.

halgari: But +1 to the idea here

pixie/vm/stdlib.py
```diff
@@ -337,16 +337,36 @@ def apply__args(args):
...
 @as_var("satisfies?")
 def _satisfies(proto, o):
-    affirm(isinstance(proto, Protocol), u"proto must be a Protocol")
-
-    return true if proto.satisfies(o.type()) else false
+    from pixie.vm.persistent_vector import PersistentVector
+    t = o.type()
+    if isinstance(proto, Protocol):
+        return true if proto.satisfies(t) else false
+    elif isinstance(proto, PersistentVector):
```
halgari: Same as with `instance?`

halgari:
Pretty epic PR. Address the few comments I posted and I'll merge it.

pixie/vm/stdlib.py
heyLu: Maybe we should introduce an internal `-instance?` (and `-satisfies?`) and then define this version in `stdlib.pxi`?

truecolour:
@heyLu That was my idea as well.

I'll update the PR this evening.

truecolour:
Actually, now that I think about it, the more common case for `satisfies?` is probably checking if a value satisfies all protocols, not any one of them. If there are no objections I'll include that change.

heyLu:
A note regarding `condp` and `case`: I think they should throw if no clause matches and there is no default expression, as they do in Clojure. If someone wants to return nil they can provide it as the default expression.

truecolour:
Here are the changes I promised.

I've got to say, working with a Lisp again is loads of fun ;-).

halgari:
Thanks!

PR #264, opened 2015-03-31
Implement fs/size and fs/permissions

justinj:


heyLu:
Can you try doing it via the FFI? It should be possible, here's a related example that also returns a struct: https://github.com/pixie-lang/pixie/blob/master/pixie/ffi-infer.pxi#L306-L321

justinj:
Ok, I'll look into that. Looks like I'll need to come up with a test that will work on Travis too.

justinj:
I'm having some difficulty getting this to work with FFI across both OSX and Linux, would someone be willing to look at this with me over IRC sometime?

halgari:
I think the problem here is how you are including c, the name libc is not the same on linux and osx. But probably we should just implement permissions via libuv. That way they are a) async, b) cross-platform. If you want to look into that, it shouldn't be that hard to copy what is in `pixie.io` and just change a few function names. Otherwise, if you don't think you'll have the time, I'll try to take care of it soon. 

Let me know either way so that we don't duplicate work.

justinj:
Ah I see, I didn't realize libc wasn't the same across linux and osx. I should be able to do it later this weekend (Sunday, probably), but if you'd like to do it sooner than that feel free.

justinj:
Ok, let me know how that looks. `defuvfsfn` is neat, by the way.

halgari:
Looks good, thanks!

PR #269, opened 2015-04-04
Package makefile

halgari:
Add release entries for ARM, x86_64 and OSX. These are pretty specific to my config, but I figure I'll be doing releases for now, so it shouldn't matter. Anyone can feel free to fix these up as desired to make them more generic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ build_preload_no_jit: fetch_externals
 	$(PYTHON) $(EXTERNALS)/pypy/rpython/bin/rpython $(COMMON_BUILD_OPTS) target_preload.py
 
 build: fetch_externals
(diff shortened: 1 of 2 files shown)


halgari:
I'm unhappy with the results of this, closing the PR until I get it fixed up a bit.

andrewchambers:
I think the current makefile is a bit odd. None of the rules are marked as .PHONY, and none really use the dependency system of make because we don't have incremental builds. The only thing which could use this is the download rules, but they don't.

I would be more in favor of having a build.py which does the same thing. We already require python for a build anyway.

MatthewWest:
@andrewchambers I'd be interested in seeing a build.py—I was under the impression that autoconf and automake were essentially the standard build system, but was dreading learning them to write a more flexible build system.

PR #281, opened 2015-04-13
Basic parser support + JSON reader

halgari:
Implements a way of building parsers using a declarative DSL. Could be improved, but the current incarnation works fairly well. Somewhat based off of OMeta: http://tinlizzie.org/ometa/


heyLu:
This looks great!

I think you want to remove the duplicated JSON parser from `pixie.parser`, though.

Also, I think the namespace for the json parser should be `pixie.parser.json`, not `pixie.parsers.json`.

heyLu:
Can you add the JSON parser back in again or are there problems with it? It might have gotten lost in the last commit.

PR #282, opened 2015-04-13
Make everything extend Object

### LINKED ISSUE #250: Should everything be an instance of Object?
This would be nice so we could use `extend-protocol` on Object.

thomasmulvaney:
Fixes #250 

I couldn't figure out a nice way to forward-declare the `Object` type. As the type class needs to be declared before any types can be. So I introduce a function `NewType` for constructing Types which wraps the the normal Type instantiater, passing in the `Object` type.

Unfortunately this means a lot of pointless `s/Type/NewType/g`ing - If you anyone can think of a better way I would be happy to modify this.


halgari:
I think this is a bit cleaner, but yeah that forward declaration is a bit tricky: https://github.com/pixie-lang/pixie/pull/283

thomasmulvaney:
That's way better. Closing this.

PR #283, opened 2015-04-14
make everything inherit from Object

### LINKED ISSUE #250: Should everything be an instance of Object?

halgari:
Fixes #250


thomasmulvaney:
Thats certainly a lot cleaner than my attempt :) . The object_inited thing is neat.

PR #285, opened 2015-04-15
Implement reify similar to Clojure

truecolour:
This PR adds a macro `reify` that creates an anonymous extension of some protocols. Because pixie encorages heavy use of polymorphism this should prove very useful.

Right now `Protocol.satisfies` (and by extension `pixie.stdlib/satisfies?`) do not work correctly on instances of `Reified`. I'll fix that soon.

Also, someone should probably give the changes in `pixie/vm/` a quick look over. I seem to suffer from RPython dyslexia, which causes the build to fail with incomprehensible error messages every time, no matter how often I triple-check the code with the PyPy coding guide.


justinj:
Whoa, this is really cool.

Random (maybe silly) question, would it be possible (or desirable) to implement this within Pixie? Seems like you could maybe have a macro that generates a `deftype` and then a single instance of that type?

halgari:
So the tough thing about reify is making it fast and handling the types. So for example: 

``` clojure

defn make-a []
  (reify 
    IFn 
    (-invoke [] :a)))

(defn make-b []
  (reify 
    IFn 
    (-invoke [] :b)))


(assert (identical? (type (make-a)) (type (make-a))))

(assert (not (identical? (type (make-a)) (type (make b)))))
```

In other words, each time a given reify is called, it should hand back an instance of the same type, but no two reifies should ever hand back the same type. Also, within pixie, this sort of thing should be possible: 

``` clojure

(extend (type (make-a))
  IObject
  (-hash [this] 42))

```

Finally whenever we add new polymorphism (like in .get_impl) in this code, we have to hit some stuff to the JIT about it (all the elidable_promote and _rev stuff). So I like the idea here, but there's still a bit of work to do. 

Probably the best first step would be to update this patch so that each reify gets its own type that we can extend and that will pass the `(not (instance? ...))` tests. 

Thanks for working on this!

halgari:
Closing as this hasn't seen progress in over two weeks, feel free to re-open if needed.

truecolour:
Sorry, didn't have the time. Not reopening since I don't know how the new VM will affect this.

PR #290, opened 2015-04-17
Tty

thomasmulvaney:


thomasmulvaney:
I added a write-stream function to pixie.io only because there wasn't something like spit or println for output-streams. Maybe there is something better we could do in this case?

pixie/system.pxi
```diff
@@ -0,0 +1,13 @@
+(ns pixie.system
```
thomasmulvaney: Calling this namespace `pixie.posix` might be a better idea?

halgari:
Can we update spit/slurp to accept streams as well as strings?

pixie/io.pxi
```diff
@@ -172,6 +172,20 @@
...
+  (let [fp (buffered-output-stream output-stream
+                                   DEFAULT-BUFFER-SIZE)]
+    (fn ([] 0)
+      ([_] (dispose! fp))
+      ([_ chr]
+       (assert (integer? chr))
+       (write-byte fp chr)
+       nil))))
+
+(defn write-stream [output-stream val]
```
halgari: This should be run through the UTF8 logic in pixie.streams.utf8. But I'm fine with that being in a different PR. I need to abstract the UTF8 stuff a bit anyways.

thomasmulvaney:
I've updated slurp and spit now to take either filenames or streams. Spitting now runs through a utf8 output stream. I haven't added utf8 streams to slurping. 
Note: Slurping TTYInputStreams breaks due to EOF errors right now.

halgari:
Merging since the EOF error is something I will work on this week. Currently we can't filter exceptions by type, so catching EOF is a bit hard. I'll try to fix that soon.

PR #292, opened 2015-04-20
Fix issue #288

### LINKED ISSUE #288: Integer remainder
Is (rem 5 2)  meant to work?  returns a double, not an integer as I had expected..

MatthewWest:
Only extend the rem operation to use math.fmod if one or both of the
two arguments is not an integer.

Fixes #288

### DIFF
--- a/pixie/vm/numbers.py
+++ b/pixie/vm/numbers.py
@@ -134,9 +134,9 @@ def define_num_ops():
                 if op == "_div" and c1 == Integer and c2 == Integer:
                     continue
                 extend_num_op(op, c1, c2, conv1, sym, conv2)
-            extend_num_op("_quot", c1, c2, conv1, "/", conv2, wrap_start = "rt.wrap(math.floor(", wrap_end = "))")
             if c1 != Integer or c2 != Integer:
                     extend_num_op("_rem", c1, c2, conv1, ",", conv2, wrap_start = "rt.wrap(math.fmod(", wrap_end = "))")
+                    extend_num_op("_quot", c1, c2, conv1, "/", conv2, wrap_start = "rt.wrap(math.floor(", wrap_end = "))")
             for (op, sym) in [("_num_eq", "=="), ("_lt", "<"), ("_gt", ">"), ("_lte", "<="), ("_gte", ">=")]:
                 extend_num_op(op, c1, c2, conv1, sym, conv2,
                               wrap_start = "true if ", wrap_end = " else false")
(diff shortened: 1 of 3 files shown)


pixie/vm/numbers.py
```diff
@@ -135,7 +135,8 @@ def define_num_ops():
                     continue
                 extend_num_op(op, c1, c2, conv1, sym, conv2)
             extend_num_op("_quot", c1, c2, conv1, "/", conv2, wrap_start = "rt.wrap(math.floor(", wrap_end = "))")
-            extend_num_op("_rem", c1, c2, conv1, ",", conv2, wrap_start = "rt.wrap(math.fmod(", wrap_end = "))")
+            if c1 != Integer or c2 != Integer:
```
heyLu: This must be an `and`, not an `or`, otherwise some operations are not defined.

I think `_quot` could be included in this `if`-statement as well, because it also has a special case defined above.

MatthewWest: It actually doesn't though. See, it's checking for whether either is not an integer. That is functionally identical to checking whether it is not the case that both are integers. Unless there's something else I'm not thinking of.

heyLu: Ah yes, sorry.

Can you include `_quot` as well, though?

MatthewWest: Done

halgari:
I'm fine with this patch, but I'd like a few tests around the results of this patch, just to verify the behavior is what we expect. I'll merge at that point.

andrewchambers:
Yeah, clearly this issue is possible because no one is writing any numeric code with pixie yet. I found it while solving project Euler problem 1.

MatthewWest:
I just added a few tests that I think handle all the possible cases, or at least most of them. All the tests are passing.

Interesting fun fact: quot returns Integer, even with fed with two ratio arguments, because it is the nearest integer quotient. Very logical!

MatthewWest:
Thanks @thomasmulvaney I fixed it to not use stringification.

tests/pixie/tests/test-numbers.pxi
```diff
@@ -49,3 +49,16 @@
 (t/deftest test-float
   (doseq [[x f] [[1 1.0] [3 3.0] [3.333 3.333] [3/2 1.5] [1/7 (/ 1.0 7.0)]]]
     (t/assert= (float x) f)))
+
+(t/deftest rem-types
+  (t/assert= "<type pixie.stdlib.Integer>" (str (type (rem 5 3))))
```
heyLu: Can you fix these asserts as well?

MatthewWest:
@heyLu Done.

halgari:
Looks great, we'll merge this as soon as the build finishes

PR #295, opened 2015-04-21
string representations for Range

stuarth:


thomasmulvaney:
This is really cool! Maybe we could make it return `"(1 2 3 ... 98 99 100)"` for really long strings?

halgari:
If anyone ever implements Tagged literals (
http://clojure.org/reader#The%20Reader--Tagged%20Literals) we could have:

(assert= (range 100) #pixie.stdlib.Range [0 100 1])

On Tue, Apr 21, 2015 at 10:43 AM, Thomas Mulvaney <notifications@github.com>
wrote:

> This is really cool! Maybe we could make it return "(1 2 3 ... 98 99 100)"
> for really long strings?
>
> —
> <https://github.com/pixie-lang/pixie/pull/295#issuecomment-94865666>.
>



-- 
programs.”
(Robert Firth)

PR #299, opened 2015-04-22
add a time macro

thomasmulvaney:


halgari:
Awesome, I've wanted this for some time. Thanks for writing it up.

thomasmulvaney:
Woke up realizing I made a boo boo. htime returns time in nanoseconds - such accuracy. But i'm only multiplying by 1000 which gives microseconds... So we actually need to multiply by 1,000,000 to get milliseconds. 

Apologies to anyone who thought pixie was 1000 times slower than it actually is ;)

Will get a PR in shortly.

PR #301, opened 2015-04-23
fix

thomasmulvaney:


jkleiser:
Looks like the right solution.

thomasmulvaney:
Thanks, @jkleiser I only realized my mistake whilst benchmarking against clojure. It's always impressive when you write code which causes a 1000x improvement in speed ;)

PR #303, opened 2015-04-23
loop destructure

stuarth:
Allow for destructuring in loop bindings, e.g.

```clj
(loop [[a b] [1 2]]
  ...)
```


halgari:
Thanks for this, it's been annoying me for some time.

PR #317, opened 2015-05-15
Refactor io

thomasmulvaney:


justinj:
I might be doing something wrong, but I'm having trouble running the tests after this. I think there's a circular dependency from `io-common -> uv -> ffi-infer -> io-blocking -> io-common`.

thomasmulvaney:
The tests all seemed to pass for me. Have you done a `make clean_pxi` then `make compile_src`?

justinj:
Hm, if I `make compile_src` first then they pass, but not otherwise. Is that expected of `pxic`s?

thomasmulvaney:
So if I'm following correctly, doing a `make clean_pxi` and then calling `make run_interactive_tests` fails?

justinj:
`make run_built_tests` fails with something like this:
```
in pixie function cb-stream-writer

in /Users/justin/dev/repos/pixie/pixie/io/common.pxi at 47:21
          uv_write (uv/uv_write_t)]
                    ^
RuntimeException: :pixie.stdlib/AssertionException Var uv_write_t is undefined
```

`make run_interpreted_tests` seems to run for a while and then fail.

Everything seems to work at ea20df6.

thomasmulvaney:
Thanks, I can confirm that this occurs for me too when I rm the `pxic` files. I'm not sure what's going on exactly. But will dig into it a bit tonight.

justinj:
Cool, thanks!

thomasmulvaney:
I can compile just io/common.pxi and get the tests to run with out errors. That seems to be the minimum amount of compilation required to make things work. The question is why... I don't see any circular dependencies so I don't think that's the issue. Very curious indeed.

Issue #318, opened 2015-05-16
Pixie fails to build with the current version of PyPy


FungusHumungus:
Currently running 

    make build_with_jit

produces the following error 

> translation:ERROR] AttributeError: \<Struct CIF_DESCRIPTION { cif, abi, nargs, rtype, atypes, exchange_size, exchange_result, exchange_args }\> instance has no field 'exchange_result_libffi'

Running OSX 10.9.5

Using an older version of PyPy works fine.

chunsj:
I have this problem. How can I solve this?

halgari:
Fixed in #321 thanks to @justinj

PR #320, opened 2015-05-19
Make comp work when the number of arguments is divisible by 4.

truecolour:
Also add some tests.


justinj:
Build failing is because of #318.

justinj:
@truecolour I think if you rebase up to 9691356 the build should pass.

PR #321, opened 2015-05-19
Update FFI for newer PyPy

justinj:
In current pypy (as of https://bitbucket.org/pypy/pypy/commits/749bf9a13d9ccf18ed9be89ac3f3ec4dd6b11a76),
exchange_result_libffi has been removed.
This means that anyone installing Pixie now and pulling down the latest
pypy will be unable to build (as in #318).

The logic for dealing with endianness has been pushed into rpython, meaning pixie/pypy don't need to deal with it.

I *think* this is all that needs to be changed, but would appreciate if someone else could take a look over the pypy commit to make sure all is good. Is there a nice way to test on different endiannesses?

People might need to download the newer pypy after this. (just `rm -rf externals` in the pixie dir).

### DIFF
--- a/pixie/vm/libs/ffi.py
+++ b/pixie/vm/libs/ffi.py
@@ -119,7 +119,7 @@ class FFIFn(object.Object):
         return exb, tokens
 
     def get_ret_val_from_buffer(self, exb):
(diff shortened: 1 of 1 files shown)


halgari:
Thank you for handling this, I've been meaning to take a look at this, but I've been really busy with life lately.

PR #322, opened 2015-05-20
keywords can now be fully qualified as in clojure

### LINKED ISSUE #316: Keywords with namespace aliases should expand
Currently we can create keywords like `::foo` but not like `::b/foo`. If `b` is an alias to `bar` in the current environment, it should expand to `::bar/foo`. If `b` is not an alias an error should be thrown.

thomasmulvaney:
Fixes #316


halgari:
Thanks for this, don't know why it's taken so long to merge

PR #324, opened 2015-05-30
Compile time checks

thomasmulvaney:
This is a bit of a big pull request and touches lots of files. It does pick up on a bunch of typos and possible bugs though and will hopefully stop these mistakes in the future. 

Basically every symbol gets checked to see if it exists in the namespace and throws an error if it can't be resolved. For example previously i could write a function `(defn foo [x] not-defined)` and there would be no compilation error.

### DIFF
--- a/pixie/streams/utf8.pxi
+++ b/pixie/streams/utf8.pxi
@@ -39,7 +39,7 @@
                             (= 0xF0 (bit-and ch 0xF8)) [(bit-and ch 7) 4 false]
                             (= 0xF8 (bit-and ch 0xF8)) [(bit-and ch 3) 5 true]
                             (= 0xFC (bit-and ch 0xFE)) [(bit-and ch 1) 6 true]
-                            :else [n 1 true])]
+                            :else [ch 1 true])]
       (loop [i (dec bytes)
              n n]
         (if (pos? i)
--- a/pixie/channels.pxi
+++ b/pixie/channels.pxi
@@ -17,6 +17,9 @@
   (-close! [this] "Closes the channel, future writes will be rejected, future reads will
                    drain the channel before returning nil."))
 
+(defn canceled? [this]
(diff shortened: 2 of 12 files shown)


pixie/streams/utf8.pxi
```diff
@@ -39,7 +39,7 @@
                             (= 0xF0 (bit-and ch 0xF8)) [(bit-and ch 7) 4 false]
                             (= 0xF8 (bit-and ch 0xF8)) [(bit-and ch 3) 5 true]
                             (= 0xFC (bit-and ch 0xFE)) [(bit-and ch 1) 6 true]
-                            :else [n 1 true])]
+                            :else [ch 1 true])]
```
thomasmulvaney: I'm not actually sure what this should be..

andrewchambers:
Does work top down like clojure? or can you define things in any order.

thomasmulvaney:
It is top down just like Clojure. Love it or hate it, it catches lots of issues. 

I'll have to figure out why the build is failing tomorrow...

andrewchambers:
I don't think this should be in the compiler. A linter maybe. This pixie is much more dynamic than clojure.

andrewchambers:
It seems like a poor man's type checker. Might as well just port core.typed.

halgari:
It does catch quite a few errors, but it is nice to be able to define code in a bottom-up manner. I've grown quite accustomed to having this flexibility, but perhaps that's just because I've spent so much time in RPython.

thomasmulvaney:
I agree that not having to follow a top down structure can be nice. This could be achieved by looking through all Vars at the end of compiling a namespace and checking that they are all mapped to something? Alternatively, we could do some kind of reference counting. When resolve_var is called we could increment a counter. Then at the end of compiling a namespace we could throw an error if any var has zero refs.

PR #330, opened 2015-06-03
BufferedInputStreams implement ISeekableStream

thomasmulvaney:
This makes buffered input streams seekable. I've tried to be efficient and only seek upstream when the position we are seeking to is outside the buffer.


halgari:
I'll merge if we can get this rebased with master

thomasmulvaney:
Rebased

PR #331, opened 2015-06-03
adds promotion to bigint on overflow

### LINKED ISSUE #58: Integer overflow handling
Currently in pixie and integer is an integer, and a big integer is a big integer...

Is Pixie planned to have transparent promotion to arbitrary precision integers like python? or is it going  to throw an exception on overflow like clojure seems to. 

Personally I think the python method is quite nice and "just works" - Really just a language design question.

thomasmulvaney:
This should fix #58

### DIFF
--- a/pixie/vm/libs/pxic/reader.py
+++ b/pixie/vm/libs/pxic/reader.py
@@ -4,7 +4,7 @@ from rpython.rlib.runicode import str_decode_utf_8
 from pixie.vm.string import String
 from pixie.vm.keyword import Keyword, keyword
 from pixie.vm.symbol import Symbol, symbol
-from pixie.vm.numbers import Integer, Float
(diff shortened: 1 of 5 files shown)


andrewchambers:
cool!

thomasmulvaney:
Unfortunately it looks like compile_tests fails at the moment because pxic_writer doesn't support BigIntegers. Will try and fix that.

MatthewWest:
Can this be merged now? I'm going through old issues to see what I can help resolve, and this looks like it could be merged.

heyLu:
Yes, we can merge this, but I don't want to decide this on my own.

@halgari Can you have a look at this?

thomasmulvaney:
Cheers, I'm going to open up writing BigIntegers (non string format) as an issue so it doesn't go forgotten..

PR #332, opened 2015-06-14
clarify parser docs, typo fix

stuarth:
`parser` doc was prob originally written for `defparser`, quick clarification to it

### DIFF
--- a/pixie/parser.pxi
+++ b/pixie/parser.pxi
@@ -191,10 +191,10 @@
       rules)))
 
 (defmacro parser
-  "(parser nm inherits & rules)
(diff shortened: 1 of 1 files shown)


pixie/parser.pxi
```diff
@@ -192,9 +192,9 @@
 
 (defmacro parser
   "(parser nm inherits & rules)
```
heyLu: Can you remove the `nm` argument from here? I think it's also a copy-paste leftover, only `defparser` seems to need that as an argument.

stuarth:
@heyLu good catch, thanks! it's updated

heyLu:
Cool, thanks!

PR #339, opened 2015-06-24
Removed circular dependency

thomasmulvaney:
uv -> ffi-infer -> io-blocking -> io.common -> uv

This is fixed by making io.common require only pixie.stream
and move all uv related io stuff into io.uv-common.

### DIFF
--- a/pixie/io-blocking.pxi
+++ b/pixie/io-blocking.pxi
@@ -2,7 +2,6 @@
   (:require [pixie.streams :as st :refer :all]
             [pixie.io.common :as common]))
 
-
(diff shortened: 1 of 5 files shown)


thomasmulvaney:
One suggestion I have is to run two tests on Travis. One before `make compile_src` and `make compile_tests` and one after as we do now. The test will be a few seconds longer but should catch any issues like this in the future.

heyLu:
Thanks!

Not sure about running the tests twice, because to be completely sure we would have to do it both with and without optimization/jit.

PR #340, opened 2015-06-24
Null pointer callbacks

thomasmulvaney:
Some C libraries will accept an optional callback and often expect a null pointer if no callback is given.

A nil callback will now be cast to a null pointer.


thomasmulvaney:
Unfortunately, this is a little awkward to write tests around without making a C library which takes a NULL pointer or a callback. I implemented this to get Zlib working, so I'm closing this and leaving it for the Zlib io PR.

PR #341, opened 2015-06-26
basic sort based on merge-sort

thomasmulvaney:
satisfies? searches up parent types


thomasmulvaney:
The `IComparable` protocol seemed like a perfect fit for `DoublePolymorphicFn`'s.

They haven't been exposed yet in regular pixie code, so I was wondering what the best way to represent such a function would be in pixie. I was thinking perhaps:

```
(extend-protocol IComparable
   [Number Number]
   (-compare [x y] ...)

  [String String]
  (-compare [x y] ...)
``` 

Any thoughts?

halgari:
I like it. Long term, I'd like to support variadic polymorphism (like Julia) but this syntax would work perfectly for something like that as well.

thomasmulvaney:
Looks like trampolining isn't working with JIT enabled. Are there some known issues with recursion in JIT mode?

Issue #319, opened 2015-05-16
Check if Boost is installed and give error if it isn't


justinj:
It would be nice if there was a nice error message telling someone they were missing it, instead of how it is now, where they get an EOF error in the middle of some FFI stuff.

heyLu:
Added in #342.

PR #342, opened 2015-06-28
#319 Add a reasonable error message if boost is not found

MatthewWest:
This is a bit of a hack implementation, I know—what should really happen is that we should switch to using autoconf and automake to check dependencies and generate build targets. Essentially it just checks two common include directories: `/usr/local/include` and `/usr/include` for the boost folder, and if it isn't there, it refuses to compile and prints a message that Boost is not found. This addresses #319 

I also updated the Readme a bit—the current size of the executable is 10.3 MB, so I updated that. Also, I used `;;` for the comments rather than `(comment ...)`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,13 @@ help:
 	@echo "make fetch_externals	   - download and unpack external deps"
 
 build_with_jit: fetch_externals
-	$(PYTHON) $(EXTERNALS)/pypy/rpython/bin/rpython $(COMMON_BUILD_OPTS) --opt=jit target.py
+	@if [ ! -d /usr/local/include/boost -a ! -d /usr/include/boost ] ; then echo "Boost C++ Library not found" && false; fi && \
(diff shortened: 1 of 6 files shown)


MatthewWest:
Added a commit to encode BigIntegers natively in compiled code (mostly copied the implementation for Integers).

andrewchambers:
Just a note, there is no good reason to use autotools or automake ever. Especially in a project like this. A python script it all that is needed and can be assumed to exist on every OS we support.

MatthewWest:
Bigintegers implemented #345

MatthewWest:
@andrewchambers I haven't seen a good example of a python script as a build script. Can you point me to an example?

andrewchambers:
Pixie does not have a complicated or incremental build, Autotools are a big mess of unnecessary complexity here.

If I have time I will try work a prototype.

heyLu:
Thanks, @MatthewWest!

PR #348, opened 2015-07-15
working zlib implementation

thomasmulvaney:
pixies ffi can convert nil to an NULL PTR callback


thomasmulvaney:
This doesn't yet expose all the functionalty of zlib  and only implements the stream version of gzip compression.

PR #352, opened 2015-07-21
Rm prompt reader

thomasmulvaney:


heyLu:
Thanks again! :sparkles:

PR #354, opened 2015-08-06
Adding a string reverse function.

cgore:


cgore:
Is that a pre-existing error?  I don't think the circleci error is related to this PR?

halgari:
Thanks for the PR!

cgore:
You’re welcome.  Thanks for the cool little language.

Christopher Mark Gore
cgore@cgore.com <mailto:cgore@cgore.com>
http://www.cgore.com <http://www.cgore.com/>

PR #356, opened 2015-08-11
Adding a split-lines function to pixie.string.

cgore:


cgore:
This seemed like the best way to split-lines with regexes.  Are there regexes in Pixie yet?

tests/pixie/tests/test-strings.pxi
```diff
@@ -27,6 +27,27 @@
...
+  (t/assert= (s/split-lines "bibbidi\nbobbidi\r\nboo")
+             ["bibbidi" "bobbidi" "boo"])
+  ;; Doesn't split lonely \r
+  (t/assert= (s/split-lines "bibbidi\nbobbidi\rboo")
+             ["bibbidi" "bobbidi\rboo"])
+  ;; Works with a single line.
+  (t/assert= (s/split-lines "BibbidiBobbidiBoo")
+             ["BibbidiBobbidiBoo"])
+  ;; Works with empty strings.
+  (t/assert= (s/split-lines "") "")
```
heyLu: The tests fail here, the expected value should be `[""]`.

heyLu:
Thanks!  No regexp support yet.  If you want to work on it you might want to have a look at how pypy does it.

cgore:
Regexes would be nice.  I'm not sure how much work that would be though.

heyLu:
It won't be trivial, that's for sure.  I tried adding it in #43 a while ago, but it turned out to be more complicated.  I haven't looked in detail at what pypy does, so maybe it is simpler than expected.

Before adding regexps to pixie you could also try to write a binding for one of the regexp libraries (PCRE, RE2, ...).

PR #361, opened 2015-08-20
Add sort support from @thomasmulvaney

halgari:
Re-adds @thomasmulvaney's merge-sort. It's not a bad algorithm, although the JIT will probably throw a fit with it. We should improve it in the future to a) use mutable arrays during sorting and b) investigate how to get the JIT to be happier. Perhaps we should re-write it in RPython to get C-like speeds and call into Pixie for the comparisons. At any rate, we need some sort of sort at this point. So let's work with this.


thomasmulvaney:
I agree its not ideal. We could probably copy rpythons listsort code and make it so a comparison function can be passed.

thomasmulvaney:
@halgari I've just implemented a better version using pixies mutable arrays and get performance on par with Clojure when run in JIT mode. It just needs some tidying up.

PR #363, opened 2015-08-22
faster-reducing repeat

joshua-g:
Hi again. I have a couple of proposed commits here.

The first thing is, my previous PR about removing promotions from custom_types.py was overzealous. Promotion is just fine in the CustomType class itself. It's only the _instance_ class where, I think, promoting "self" was harmful to performance. Sorry! I'm still trying to learn all of this. Adding back in these promotions makes field access faster in Range's reduce loop.

The second proposed commit is a deftype for Repeat. It's analogous to the one we have for Range. The main reason I think this is a good idea: fast reducing/transducing over repeats without the lazy-seq overhead. I think that's important since repeats are some of the basic building block sequences that should be as fast as possible.

To be consistent with how Range works, -seq generates a plain old lazy-seq.

(This actually came up because I was trying to test a SHA-3 implementation I'm working on in pixie; one of the test vectors is one million repeats of the letter 'a').

After this change, `(reduce + (repeat 1000000000 1))` finishes for me in less than a second, as does `(reduce + (range 1000000000))`.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1347,12 +1347,6 @@ and implements IAssociative, ILookup and IObject."
   (puts (apply pr-str args))
   nil)
 
-(defn repeat
-  ([x]
-     (cons x (lazy-seq* (fn [] (repeat x)))))
-  ([n x]
-     (take n (repeat x))))
-
 (defn repeatedly
   {:doc "Returns a lazy seq that contains the return values of repeated calls to f.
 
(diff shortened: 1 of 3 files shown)


heyLu:
Thanks so much for this!  (And sorry for getting back so late.)

Also thanks for the detailed descriptions, these really help!  (You might want to include some of that information in the extended commit message as well in the future, that way it's easier to read when going through the log or individual commits.)

joshua-g:
@heyLu That's a good idea. :) Will do.

PR #364, opened 2015-08-28
fixed (count (lazy-seq [])) incorrectly returning 1

joshua-g:
stdlib.py's count was incrementing acc before checking if the current object is empty. It assumed that all empty collections satisfy ICounted, but lazy-seq does not.

Also I added some tests for count.


heyLu:
Thanks!

PR #365, opened 2015-08-31
udp client + server

stuarth:
first cut at adding udp. what do we think?

### DIFF
--- a/pixie/io/ip.pxi
+++ b/pixie/io/ip.pxi
@@ -1,0 +1,11 @@
+(ns pixie.io.ip
+  (:require [pixie.uv :as uv]))
+
+(defn ip4-sock-addr
+  [ip port]
+  (assert (string? ip) "Ip should be a string")
+  (assert (integer? port) "Port should be a int")
+  
+  (let [addr (uv/sockaddr_in)]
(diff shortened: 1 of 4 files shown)


heyLu:
Thanks for working on this!

Could we have an interface that's similar to the tcp one?  In particular, I like that it returns a stream that you can read from and write to.

halgari:
Does a stream make sense with a message based protocol like UDP?

heyLu:
Apparently, not so much.  The last time I looked I *thought* various languages provided stream-like APIs to UDP.  Some do, but they all *also* provide a packet/message oriented API, which is likely the preferred one.

Not sure what the API should look like in Pixie, though.

(Examples: [Ruby][], [Go][], [Java][])

[Ruby]: http://ruby-doc.org/stdlib-1.9.3/libdoc/socket/rdoc/UDPSocket.html
[Go]: https://golang.org/pkg/net/#UDPConn
[Java]: https://docs.oracle.com/javase/8/docs/api/java/net/DatagramSocket.html

stuarth:
Closing until I have a chance to revisit. Fair game for anyone to reopen + rework :)

PR #370, opened 2015-09-09
First stab at implementing ns-publics

arrdem:
Not quite sure if I'm accessing the metadata correctly as I'm still wrestling the pypy build process on my dev box but this is at least a nearly working sketch.

### DIFF
--- a/pixie/vm/stdlib.py
+++ b/pixie/vm/stdlib.py
@@ -589,6 +589,38 @@ def ns_aliases(ns):
 
     return nil
 
+@as_var("is-public?")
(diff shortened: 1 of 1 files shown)


heyLu:
Mhh..., why did you do this in python? I think it should be possible to do in pixie itself, by using `ns-map`.

arrdem:
Agreed. I'll follow up on this later this week or next.

Some work to get Pixie/PyPy building locally in a mainly py3 environment will have to be done first and the first round of midterms are here :|

PR #372, opened 2015-09-10
improvements to jit and batch mode

halgari:
Some stuff I wanted for my upcoming talk at Strangeloop. Improves the performance of `float` and `int` by several orders of magnitude.

### DIFF
--- a/target.py
+++ b/target.py
@@ -39,7 +39,7 @@ PROGRAM_ARGUMENTS.set_root(nil)
 
 LOAD_PATHS = intern_var(u"pixie.stdlib", u"load-paths")
 LOAD_PATHS.set_root(nil)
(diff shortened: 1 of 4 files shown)


target.py
```diff
@@ -81,7 +83,9 @@ def inner_invoke(self, args):
                     if not path.isfile(self._file):
                         print "Error: Cannot open '" + self._file + "'"
                         os._exit(1)
-                    f = open(self._file)
+                    load_file.invoke([rt.wrap(self._file)])
```
heyLu: Do we need to catch exceptions here?

heyLu:
This pr also makes running `.pxic` files directly possible.  (I.e. `pixie-vm my-ns.pxic` will work.)

halgari:
I'll add some more error checking during the next few days.

PR #373, opened 2015-09-16
fn pre and post conditions

stuarth:
pre and post conditions in the same style as Clojure

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2154,12 +2154,12 @@ The params can be destructuring bindings, see `(doc let)` for details."}
                                        `((let [~'% ~(if (> (count body) 1)
                                                       `(do ~@body)
                                                       (first body))]
-                                           ~@(map (fn* [c] `(assert ~c)) post)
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
Perhaps the build has failed because pypy has changed?

thomasmulvaney:
I restarted the build and it's all green now. Must have pulled down a bad pypy build.

stuarth:
thanks for that @thomasmulvaney

thomasmulvaney:
@stuarth this would be really cool to get in. My only suggestion is that the assertions perhaps don't throw enough information as to where the error happened. Perhaps adding a message to the assertion such as "in precondition (even? x)" might be handy for debugging?

arrdem:
A possibility worth exploring... you could easily extend the pre/post vector grammar to be `(list string?)*` where the string following a predicate sexpr is interpreted to be the error message to be thrown should that guard fail. This is an improvement to Clojure's pre/post I've kicked around before.

heyLu:
This looks good, I'll merge this later today if nobody complains.

heyLu:
Thanks @stuarth! :sparkles:

PR #374, opened 2015-09-20
Line reader

thomasmulvaney:
This commit improves the performance of `line-seq` by introducing a reducible `LineReader`.

To keep things simple, all line operations, `line-seq`, `read-line` and `line-reader` all require either a `BufferedInputStream` or `UTF8InputStream`.

`BufferedInputStream` is now also reducible with the intention of using `(transduce (take-while #(not (line-feed? %)) string-builder reducible-stream)` instead of loops to read lines from both `UTF8InputStreams` and `BufferedInputStreams`, but the performance was terrible (~20x worse) for some reason. This is probably worth investigating some more.

`line-seq` and `-reduce` on LineReader both churn through a `BufferedInputStream` of `/usr/dict/words` in ~300ms on my machine compared to the previous ~7 seconds. The equivalent `UTF8InputStream` takes about 1 s to read now.

### DIFF
--- a/benchmarks/read-line.pxi
+++ b/benchmarks/read-line.pxi
@@ -6,17 +6,37 @@
 (def file-name "/usr/share/dict/words")
 
 (println "Lazy line-seq")
-(time/time (-> file-name
-               (io/open-read)
-               (io/line-seq)
-               (count)
-               (println)))
+(time/time 
+  (->> file-name
+       (io/open-read)
+       (io/buffered-input-stream)
+       (io/line-seq)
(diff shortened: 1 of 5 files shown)


heyLu:
Is this ready to be merged?  I'm not the best to judge this, but this looks good to me!

thomasmulvaney:
Hmm. I don't know why the travis tests didn't run. Let me rebase this and hopefully they fire.

heyLu:
Thanks a lot!

PR #378, opened 2015-09-23
fix first on Seqables

### LINKED ISSUE #376: (first {:some :map}) throws RuntimeException
It seems as though `(first {:some :map})` is intended to fall down to https://github.com/pixie-lang/pixie/blob/master/pixie/vm/stdlib.py#L135 which would be similar (equivalent?) to `(first (seq {:some :map}))`

This is the 

```clj
user => (first {:some :map})
ERROR: 
 in pixie function repl_fn

in /home/bja/code/pixie/pixie/repl.pxi at 24:24
               (let [x (eval form)]
                       ^
in internal function eval

in <unknown> at 2:1
(first {:some :map})
^
in pixie function first

in /home/bja/code/pixie/pixie/stdlib.pxi at 689:5
    (-first coll)))
    ^
in polymorphic function -first dispatching on pixie.stdlib.PersistentHashMap

RuntimeException: :pixie.stdlib/AssertionException No override for pixie.stdlib.PersistentHashMap on -first in protocol ISeq
```

thomasmulvaney:
Fixes #376

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -940,6 +940,11 @@ If further arguments are passed, invokes the method named by symbol, passing the
                                   (cond (= idx 0) (-key self)
                                         (= idx 1) (-val self)
                                         :else not-found)))
+(extend -eq MapEntry (fn [self other]
+                       (and (= (-key self)
+                               (-key other))
+                            (= (-val self)
+                               (-val other)))))
 
 (extend -reduce MapEntry indexed-reduce)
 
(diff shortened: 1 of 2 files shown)


emidln:
Does this need to extend to second, third, and fourth as well?

thomasmulvaney:
Good point. I also realized that its possible to make them all work. In pixie a protocol can extend another protocol. So I've just said that ISeq extends ISeqable.

heyLu:
Looks good!  Can you add a test for a non-empty map as well?

thomasmulvaney:
Done. This required getting MapEntry equality sorted too.

heyLu:
Thanks!

PR #379, opened 2015-09-23
readline/add_history to repl

stuarth:
`add_history` to repl


thomasmulvaney:
This looks really handy, as currently I think one has to use rl-wrap or `dust repl` to get this kind of functionality.

halgari:
Awesome, thanks!

PR #380, opened 2015-09-24
Map Entry equality

emidln:
This extends -eq on MapEntry to include vectors (as in clojure) as well as extends -seq to MapEntry so that the reverse -eq (vector to MapEntry) can also be true.

Now the following holds,

```clj
(= (map-entry :a 1) [:a 1])
(= [:a 1] (map-entry :a 1))
```

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -919,10 +919,14 @@ If further arguments are passed, invokes the method named by symbol, passing the
                                         (= idx 1) (-val self)
                                         :else not-found)))
 (extend -eq MapEntry (fn [self other]
-                       (and (= (-key self)
-                               (-key other))
-                            (= (-val self)
-                               (-val other)))))
+                       (cond
+                         (map-entry? other) (and (= (-key self)
+                                                    (-key other))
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
This looks good however whilst `(= '(:a 1) (map-entry :a 1))` is true. `(= (map-entry :a 1) '(:a 1))` is false.

emidln:
Rather than `(cond .... :else false)`, should MapEntry have `(cond .... :else (= (seq self) other))` such that we just defer to seq equality?

emidln:
I added the test for (= (map-entry :a 1) '(:a 1)) and added a fallback to calling seq on the map entry and then using = on that and other. This is perhaps not ideal, but until we can dispatch on both types for the -eq protocol, I think it's the best that can be done.

halgari:
I'm okay with merging this, but we have a merge conflict, if we can get that fixed, I'll push the button.

emidln:
I rebased against master to fix the merge conflict. The tests fail, but they're not tests for *this* code (and they fail in master too).

heyLu:
Merged, thanks!

PR #384, opened 2015-10-02
Added ns-publics plus 0 arity ns-map

emidln:
Also demoted ns-map from vm/stdlib to stdlib to add an additional 0 arity.

This is a version of ns-publics implemented in terms of ns-map. This also
adds convenience arities to both ns-map and ns-publics (defaults to *ns* 
with no arguments).


heyLu:
Hi, thanks for this!  Can you add some tests?  (I checked that it works locally, but otherwise we won't know if it breaks in the future.)

PR #385, opened 2015-10-02
Adding a pixie.string/trim-newline function based off the one in Clojure.

cgore:


heyLu:
Thanks for the PR! I think we should do this in pixie itself, though, it seems [simple enough](https://github.com/clojure/clojure/blob/d7175e84e27816ee227220023a5411386d6aa32e/src/clj/clojure/string.clj#L272).

cgore:
Doing it in native Pixie would be okay, I just did it in RPython because the other trim functions were done that way.  If I do trim-newlines in Pixie, would it make sense to move the others over to the Pixie side of the house too?

heyLu:
Yeah, that makes sense. I think we want to do as much as possible in Pixie, and `trim-newline` can be done entirely in Pixie. (The others use stuff from rpython, like `.find` or the `unicodedb` module.)

I think the others can remain in rpython, I don't see how we convert them. (Except maybe `substring` and `index-of`, both of which could be based on one internal function and the other arities could be defined in Pixie. If you want to do that, it would be nice if you'd submit a separate pr for that, though.)

cgore:
Okay, cool, I'll rewrite it in Pixie then.

heyLu:
Thank you!

PR #386, opened 2015-10-02
Adding a set union function.

cgore:


heyLu:
Thanks for this!  I'm not sure if we have to rewrite `pixie.set` from scratch though, because we could port `clojure.set` over...  Does anyone know if that would be allowed, considering that Clojure has a different license?

(I ported `clojure.set` a few months ago, but didn't submit a pull request then because of the potential licensing issues... The branch is [here](https://github.com/pixie-lang/pixie/compare/master...heyLu:set-and-walk), if anyone wants to take a look. Take it if you want to.)

cgore:
**[Important caveat: IANAL.]**

My understanding is that the Apache 2.0 and the LGPL are **NOT** compatible in that sense, the LGPL has some extra wording around keeping the downstream modifiable that is absent in the Apache 2.0 License.  Also, while it is totally cool to dual-license code, it would be something the copyright holders would have to do, we couldn't change the license ourselves, and this would apply to any license scheme, not just Apache 2.0/LGPL.

I think it would be okay from a legal standpoint if the derivative work (that is, pixie.set) remained with the Apache 2.0 license (including any future improvements to it), if the rest of Pixie remained LGPL (the two licenses can co-exist in a single project), but that would be quite confusing.

In short, my understanding is Apache 2.0 licensed code and LGPL licensed code can live side-by-side in the same code base, but the two shall never intermix, unless the original authors of the code are willing to dual-license one side or the other side, and then it all really just becomes licensed that one way for all real purposes.  And I doubt we'd get Rich Hickey to re-license his stuff for us.

**[Again, I reiterate: IANAL.]**

uvtc:
The [fsf various licenses page](http://www.gnu.org/licenses/license-list.html) indicates that Apache 2 is indeed GPL-compatible, but that Clojure's license (the Eclipse 1.0) is not.

halgari:
Set functions are easy enough to write from scratch, and Clojure's set
functions have some caveats (they accept non-set collections but don't work
properly in those cases). So I suggest we write our own.

And yes, we should stay away from copy-and-pasting Clojure EPL licensed
code directly into Pixie libraries.

Timothy

On Mon, Oct 5, 2015 at 12:19 PM, John Gabriele <notifications@github.com>
wrote:

> The fsf various licenses page
> <http://www.gnu.org/licenses/license-list.html> indicates that Apache 2
> is indeed GPL-compatible, but that Clojure's license (the Eclipse 1.0) is
> not.
>
> —
> <https://github.com/pixie-lang/pixie/pull/386#issuecomment-145589252>.
>



-- 
programs.”
(Robert Firth)

cgore:
@halgari - Would we want to add checks to ensure it is a set and reject if not, or would you prefer that we accept non-sets and just _"do the right thing"_ for non-sets?

cgore:
After thinking about my question for a bit, it'll often not be obvious what _"the right thing"_ is, even in simple cases, so it's probably better to just have the function complain loudly if the inputs aren't sets, so that's what I'll do here.

halgari:
agreed

On Tue, Oct 6, 2015 at 1:22 PM, Christopher Mark Gore <
notifications@github.com> wrote:

> After thinking about my question for a bit, it'll often not be obvious
> what *"the right thing"* is, even in simple cases, so it's probably
> better to just have the function complain loudly if the inputs aren't sets,
> so that's what I'll do here.
>
> —
> <https://github.com/pixie-lang/pixie/pull/386#issuecomment-145971273>.
>



-- 
programs.”
(Robert Firth)

heyLu:
Thanks @cgore!

Issue #203, opened 2015-03-01
Syntax quoting for maps is broken


halgari:
This doesn't work because I never took the time to implement it, we can probably just port what Clojure does for maps. 

````clojure

`{:foo ~bar}

```

PR #387, opened 2015-10-03
Fix syntax quoting for maps

nooga:
This fixes issue #203.
I've added a test for it in `test-macros.pxi` (couldn't find any existing file to stick it in).

This also adds `map?` fn, which is similar to `vector?` etc.

### DIFF
--- a/pixie/vm/reader.py
+++ b/pixie/vm/reader.py
@@ -430,6 +430,7 @@ APPLY = symbol(u"apply")
 CONCAT = symbol(u"concat")
 SEQ = symbol(u"seq")
 LIST = symbol(u"list")
+HASHMAP = symbol(u"hashmap")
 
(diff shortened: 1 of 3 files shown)


heyLu:
Congrats on your first pull request! :)

`map?` is also defined in `stdlib.pxi`, does this pr also work without defining `map?` in python?

nooga:
Thanks!

The python `map?` is only used here: https://github.com/pixie-lang/pixie/pull/387/files#diff-92073f3cf3167d94528a42b45350e025R481

I guess I could just use `rt._satisfies_QMARK_(rt.IMap.deref(), ...)` instead. 
Decided to define `map?` to make my code look more like its surroundings.

heyLu:
Merged. Thanks!

PR #388, opened 2015-10-04
Support IInputStream for line-reader

justinj:
Not sure if this is the best way to solve this, possibly nicer would be to just pass things into `buffered-input-stream` in the catchall and have that handle the errors.

cc @thomasmulvaney because I think he wrote a lot of the line-reader stuff

### DIFF
--- a/pixie/io.pxi
+++ b/pixie/io.pxi
@@ -87,8 +87,11 @@
     (instance? BufferedInputStream input-stream)
     (-> input-stream ->LineReader)
 
+    (satisfies? IInputStream input-stream)
+    (-> input-stream buffered-input-stream ->LineReader)
+
     :else
-    (throw [::Exception "Expected a LineReader, UTF8InputStream, or BufferedInputStream"])))
+    (throw [::Exception "Expected a LineReader, UTF8InputStream, BufferedInputStream, or IInputStream"])))
 
 (defn read-line
   "Read one line from input-stream for each invocation.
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
The reason I didn't do exactly this is that it introduces slightly confusing behaviour. If you take an input-stream and pass it to line-reader (as per PR) and read a line, the user would expect the input-stream to now be pointing to the next line. However this is not the case as a buffer has been read, not a line.

I don't have time to check this PR out now, but I think if you add a few more tests the issue I'm getting at might show up. Perhaps if you copy the `test-seek` test and make it use an unbuffered stream the issue may become apparent.

justinj:
Ah yeah, I see what you mean, that's annoying. Is there a nice way to handle things like `TCPStream`s when you can't call `read-line` (or anything in that family) on them because of this?

thomasmulvaney:
I think we could make a UnbufferedLineReader and line-reader could build
the appropriate `*LineReaderbased` on the input stream. I partly
implemented this when I made the original PR but left it out as using
buffered-streams is so much faster. Even though the UnbufferedLineReader
should technically be faster, reading from unbuffered streams was the
bottleneck, so didn't actually give any performance improvements. You are
right though, we do need a way to read lines on TCPStreams and TTYStreams
so I would recommend writing an UnbufferedLineReader.

justinj:
Hey, sorry for the delay, going to close this for now until I get some time to really look into it. If someone else wants to take over feel free.

thomasmulvaney:
@justinj, I think this would actually work if you set the buffer size to 1.

thomasmulvaney:
Closing this as implemented #453

PR #389, opened 2015-10-07
Adding an implementation of set intersection.

cgore:


thomasmulvaney:
Have you tried using:

`(set (filter #(contains? set-a %) set-b))`

or

`(transduce (filter #(contains? set-a %)) conj #{} set-b)`

rather than using an atom. I'm just curious about the performance.

cgore:
That is easier to read, thanks!

halgari:
Looks good, thanks!

PR #391, opened 2015-10-08
Fix keywords beginning with numbers.

### LINKED ISSUE #294: Keywords beginning with Number
I didn't find a language spec anywhere so I'm not sure if this is a bug, but in Clojure you can have keywords which are a number. I was surprised to find that in Pixie this breaks the reader.

```
user => ({:3 3 :4 4} :3)
ERROR:
 in pixie function repl_fn

in pixie/repl.pxi at 21:23
      (try (let [form (read rdr false)]
                      ^
in internal function read

RuntimeException: Can't keyword quote a non-symbol

3
ERROR:
 in pixie function repl_fn

in pixie/repl.pxi at 21:23
      (try (let [form (read rdr false)]
                      ^
in internal function read

RuntimeException: Can't keyword quote a non-symbol

4
ERROR:
 in pixie function repl_fn

in pixie/repl.pxi at 21:23
      (try (let [form (read rdr false)]
                      ^
in internal function read

RuntimeException: Unmatched Map brace '}'

ERROR:
 in pixie function repl_fn

in pixie/repl.pxi at 21:23
      (try (let [form (read rdr false)]
                      ^
in internal function read

RuntimeException: Can't keyword quote a non-symbol

ERROR:
 in pixie function repl_fn

in pixie/repl.pxi at 21:23
      (try (let [form (read rdr false)]
                      ^
in internal function read

in <unknown> at 20:16
({:3 3 :4 4} :3)
               ^
RuntimeException: Unmatched list close ')'
```

This isn't so great. And the stack trace is pretty enormous for what it is...

harrisi:
Fixes #294.


halgari:
This makes numbers implement INamed. Not sure if I agree with that. Might be a better idea to update the reader to call `rt.str` if the value is not named. Or even better, make the keyword reader call the symbol reader directly, but without interpreting the symbol. That way `:1.00000` doesn't become `:1.0`

harrisi:
Thank you for the feedback. Obviously this was the quickest and clearest way to get it "working." I'll look into what you've suggested.

thomasmulvaney:
@harrisi Thanks for working on this. I've got a PR which deals with this in #505 so I've closed this. It uses the symbol reader instead.

PR #392, opened 2015-10-08
Set predicates

cgore:
Adding the following to `pixie.set`:

- `subset?`
- `superset?`
- `strict-subset?`
- `strict-superset?`

And some basic tests for them.

### DIFF
--- a/pixie/set.pxi
+++ b/pixie/set.pxi
@@ -63,8 +63,9 @@
 (defn strict-subset?
   "Returns true if the first set is a subset of the second."
   [s t]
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
Ace! My only suggestion would be replacing the `(reduce conj set-a set-b)` with `(into set-a set-b)`  in the `union` function. This will use transient sets which should be faster.

cgore:
Cool.  For some reason I always forget that `into` is a thing.

thomasmulvaney:
Thanks!

PR #394, opened 2015-10-10
Use python2 as the default python executable

### LINKED ISSUE #337: `python` may not refer to python2 on some linux distributions.
`Makefile` assumes python refers to python 2.x, which is not true on archlinux.
Thus, it leads to errors on archlinux, and I had to set PYTHON to python2 in `Makefile`.

heyLu:
This should make building on systems with python3 as default easier, and
also make it clear that we need python2.

Fixes #337.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ all: help
 
 EXTERNALS=externals
 
-PYTHON ?= python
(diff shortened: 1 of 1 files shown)


arrdem:
:+1: awesome thanks!

PR #397, opened 2015-10-11
Fix make clean_pxic 'rm: missing operand' error

puffybsd:
`make clean_pxic` fails if no files are found due to invalid
rm command 'rm: missing operand'. Add `--no-run-if-empty` to xarg
to fix.


thomasmulvaney:
Thanks!

Issue #377, opened 2015-09-23
FFI doesn't deal with ld scripts


skrat:
So a fresh clone of pixie, when being built, will fail to compile stdlib, because on some systems (arch linux), `libm.so` is a ld script instead of ELF binary.

This might be relevant: http://www.gossamer-threads.com/lists/python/dev/941581

heyLu:
I had just happen to me as well.  Not sure how to handle it, though.  We could add a helper similar to the one for libc, but maybe we *should* handle it properly.

Here's how we find the correct libc location:

- [pixie.platform](https://github.com/pixie-lang/pixie/blob/master/pixie/vm/libs/platform.py#L12) uses `get_libc_name` from `rpython.rlib.clibffi`
- [rpython.rlib.clibffi](https://bitbucket.org/pypy/pypy/src/3c4e18e306563b898af58b332852c0c8e67ff758/rpython/rlib/clibffi.py?fileviewer=file-view-default#clibffi.py-289) uses `ctypes.util.find_library`
- and finally, `find_library` does some [dark magic involving gcc](https://hg.python.org/cpython/file/8f91b475c2fd/Lib/ctypes/util.py#l96)
    - which seems to boil down to executing `gcc -Wl,-t -o $tmpfile -l$library_name` and grepping for the full path in the result
    - we can't use `ctypes.util.find_library` (I think), because it's code from python.  we *can* use it during translation, but not at runtime

That's all I've got!  If anyone wants to do this, or has other ideas for *how* to do it, please say so.

heyLu:
The (dirty) fix is to hard code the name of `libm` in `stdlib.pxi`:

```diff
diff --git a/pixie/stdlib.pxi b/pixie/stdlib.pxi
index 880efd4..6a82bb7 100644
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -19,7 +19,7 @@
 (def fopen (ffi-fn libc "fopen" [CCharP CCharP] CVoidP))
 (def fread (ffi-fn libc "fread" [CVoidP CInt CInt CVoidP] CInt))
 
-(def libm (ffi-library (str "libm." pixie.platform/so-ext)))
+(def libm (ffi-library (str "libm.so.6" #_pixie.platform/so-ext)))
 (def atan2 (ffi-fn libm "atan2" [CDouble CDouble] CDouble))
 (def floor (ffi-fn libm "floor" [CDouble] CDouble))
[...]

PR #399, opened 2015-10-15
FFI: coerce numbers to double if required and possible

12foo:
If a function invoked over FFI requires a double and a (Big)Integer or a
Ratio is supplied, automatically coerce to double.

Should help with the rest of #377.


thomasmulvaney:
Looks good, but seeing as `to_float` is defined in pixie.vm.numbers it might be nicer to just import that and use `val = to_float(val)`  in `ffi_set_value`?

12foo:
I was wondering about that (I added Ratio to to_float for that reason), but somehow didn't do it in the end. Brain hiccup, probably.

thomasmulvaney:
I think you probably don't even need the if statement with type checks. The to_float call should raise an error about coercion which is what happens in all the other `ffi_set_value` tests.

thomasmulvaney:
Ace! Thanks for working on this!

PR #402, opened 2015-10-16
fixes getting docs from aliased namespace 

### LINKED ISSUE #401: `doc` unable to find docs for pixie.io/spit
~~~
$ pixie-vm 
user => (require pixie.io :as io)
nil
user => (doc io/spit)
ERROR: 
 in pixie function repl_fn

in pixie/repl.pxi at 27:24
               (let [x (eval form)]
                       ^
in internal function eval

in pixie function doc

in internal function resolve-in

RuntimeException: :pixie.stdlib/AssertionException Unable to resolve namespace: io inside namespace pixie.stdlib
~~~

thomasmulvaney:
Fixes #401 

Would be nice to get some better tests around doc stuff any thoughts on best way to do this?

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1371,7 +1371,7 @@ and implements IAssociative, ILookup and IObject."
    :added "0.1"}
   [v]
 
-  (let [vr (resolve v)
+  (let [vr (resolve-in *ns* v)
         x (if vr @vr)
         doc (get (meta x) :doc)
         has-doc? (if doc true (get (meta x) :signatures))]


heyLu:
Neat! :)

We could execute something like `pixie-vm -e '(require pixie.io/spit :as io) (doc io/spit)'` in the tests, and check that it exits successfully.  (We need some tests for `pixie-vm` anyway, so maybe just add it there.)

heyLu:
Thanks!  (We still might want to add tests, but hopefully we'll have a nicer idea of how to do that later.)

thomasmulvaney:
I agree tests around calling `pixie` itself make sense.

PR #404, opened 2015-10-19
Add coll? function

alejandrodob:
Just added the coll? function.

Some things to notice:

+  `(coll? nil)` evaluates to `true` while in Clojure it evaluates to `false`. Not sure about what the expected behaviour should be.
+ I was also not sure about whether transient collections should return `true` or `false` when being evaluated with `coll?`. This PR mimics Clojure in that, so `(coll? (transient [1]))` returns `false`.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -867,6 +867,7 @@ If further arguments are passed, invokes the method named by symbol, passing the
 (defn set? [v] (instance? PersistentHashSet v))
 (defn map? [v] (satisfies? IMap v))
 (defn fn? [v] (satisfies? IFn v))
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
Looks good. I think I agree with transients returning `false`. If you are checking whether something is a collection so you can call `conj` then only `IPersistentCollections` make sense.

Following the above train of thought, if we call `(conj nil 1)` in clojure we get `(1)` which seems a little inconsistent given `(coll? nil)` returns `false`. So I personally feel like nil should return `true` as `nil` is `conj`able.

thomasmulvaney:
Thanks!

PR #407, opened 2015-10-27
add vary-meta

mpenet:


mpenet:
travis is not building the right commit :[ 
If you could manually relaunch it, or maybe I can just do a "touch" commit to trigger that.

mpenet:
I ll reopen a new PR, travis doesn't seem to be able to test it correctly (I am fairly sure the test should pass).

PR #408, opened 2015-10-27
add IRecord marker protocol and record? predicate

mpenet:
I also had to move string-builder otherwise it wouldn't compile


heyLu:
Thanks again! :)

PR #411, opened 2015-10-28
add metadata support to records

mpenet:
It should be fairly safe, I just added a __meta field to the underlying deftype and then overwrite the constructor to set the metadata field to nil. 

In the process I also noticed that records do not currently permit to enumerate their keys / values via `keys` `vals`, but for now I guess it makes things more simple.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1281,6 +1281,7 @@ and implements IAssociative, ILookup and IObject."
         fields (transduce (map (comp keyword name)) conj field-syms)
         type-from-map `(defn ~map-ctor-name [m]
                          (apply ~ctor-name (map #(get m %) ~fields)))
+        meta-gs (gensym "meta")
         default-bodies ['IAssociative
                         (-make-record-assoc-body ctor-name fields)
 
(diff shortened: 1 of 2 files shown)


pixie/stdlib.pxi
```diff
@@ -1304,9 +1312,13 @@ and implements IAssociative, ILookup and IObject."
                                           fields)))
                         `(-hash [self]
                                 (hash [~@field-syms]))]
-        deftype-decl `(deftype ~nm ~fields ~@default-bodies ~@body)]
+        deftype-decl `(deftype ~nm ~(conj fields '__meta) ~@default-bodies ~@body)
+        ctor `(defn ~ctor-name ~field-syms
```
heyLu: Is this separate constructor necessary?  The `deftype` call already defines it, so we would overwrite that here...

mpenet: The constructor from the deftype includes the __meta field, so we need to "overwrite" it to allow construction of the record with __meta initalized to nil (or we get a wrong number of arg error).

mpenet: Just to expand on that, I didn't do it in deftype, because types should be able to choose not to implement IMeta (like in clojure), and be more lightweight this way (it's one less field after all).

mpenet:
Anything blocking this PR or changes required? I need it for a lib I am working on (I can always run a locally patched version of pixie, but if I can avoid it).

heyLu:
No, we're just all a bit busy.

If you can rebase this PR onto master I'll merge it.

mpenet:
Done, thanks.

heyLu:
Thanks for the PR.  Have fun! :)

mpenet:
Awesome thanks! more to come hopefully ;]

PR #415, opened 2015-10-30
Implement IReduce on Records

mpenet:
This makes select-keys/vals/keys & all work on records among other things.


thomasmulvaney:
Great stuff! Thanks!

PR #416, opened 2015-10-30
allows kw/map/set invocations with default value

mpenet:
this allows default values on -val-at call with invoke such as:

```clojure
(:x {} :bar)
(:x #{} :bar)
({} :x :bar)
(#{} :x :bar) 
```


heyLu:
Thanks again! :)

PR #423, opened 2015-11-02
check for correct EOF value - fix #413

### LINKED ISSUE #413: wrong return value check of fputc infile-output-rf  (via  write-byte)
https://github.com/pixie-lang/pixie/blob/master/pixie/io-blocking.pxi#L88 
I noticed that while copying image files around, some of the files would be corrupted because of this. 

I believe EOF should be -1 not 0 ? (if I change the line indicated for -1 this works as expected).

mpenet:


heyLu:
:sparkles: (Thanks! :)

Issue #406, opened 2015-10-25
No unchecked operations for +, -, *


reutermj:
Discussed October 22 on Gitter with Andrew Chambers.

The default behaviour of the +, -, and * math operations (to convert to BigInteger on overflow) arguably makes more since than the cyclical behaviour found in many other languages. But, there are many algorithms like the XorShift PRNG algorithm and applications like games that the cyclical behaviour makes more since or are requirements.

Having a secondary library function along the lines of u+, u-, and u* to allow for the cyclical behaviour while maintaining the default behaviour would be appreciated greatly for a lot of the applications for which I want to be using Pixie (i.e. procedural generation and games).

I have an an initial implementation of u+, u-, and u* in https://github.com/reutermj/pixie. This implementation currently works for pixie.stdlib.Integer and pixie.stdlib.Float. Currently the there are issues with the usages of pixie.stdlib.BigInteger and pixie.stdlib.Ratio. But, I can fix those, and submit a pull request if my proposed function naming and implementation of the unchecked operations are accepted.

thomasmulvaney:
Implemented in #426

PR #426, opened 2015-11-02
Added Unchecked Math Operations

reutermj:
Related to issue #406. Added operations u+, u-, and u* for Integer, Float, BigInteger, and Ratio that follows the semantics of the checked operations except for cases of Integer and Integer. In these cases, the operations do not promote to BigInteger, but instead wrap like the equivalent C/C++ operations.

I also defined tests that mirror the tests for the checked operations, as well as added a test demonstrating the wrapping behaviour.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -510,6 +510,30 @@ returns true"
   ([x y & args]
       (reduce -mul (-mul x y) args)))
 
+(defn unchecked-add
+  {:doc "Adds the arguments, returning 0 if no arguments"
+   :signatures [[& args]]
+   :added "0.1"}
+  ([] 0)
+  ([x] x)
+  ([x y] (-unchecked-add x y))
+  ([x y & args]
+      (reduce -unchecked-add (-unchecked-add x y) args)))
+
+(defn unchecked-subtract
+  ([] 0)
+  ([x] (-unchecked-subtract 0 x))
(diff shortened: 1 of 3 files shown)


heyLu:
I'm a bit unsure about this.  Here are a few comments:

- I think the names should mirror the ones in Clojure, i.e. `unchecked-add`, ...
- we may want to add them to a separate `IUncheckedMath` protocol
- I think the operations should only work on integers and floats, not ratios

I may be wrong on these, feel free to tell me if that's the case!

reutermj:
I've made changes with respect to all of your comments. To be completely honest I only knew about clojure's *unchecked-math* and didn't know of the specific functions. I also added unchecked-inc and unchecked-dec to reflect the api.

thomasmulvaney:
Ace. If you can squash the commits I'll merge this shortly. Thanks!

reutermj:
Done. Let me know if you have any problems with it.

heyLu:
Thanks!  Can you rebase this onto current master, so that the other commits disappear?

reutermj:
Sorry I thought I did that; git is still magic to me. It should be done now.

heyLu:
No worries, git tends to be confusing.  :)

Thanks a lot for the pull request, will merge when the tests have run.

heyLu:
:sparkles: :)

PR #428, opened 2015-11-03
add pixie.walk

mpenet:
It's very similar to clojure.walk but uses transients when possible, and is extensible via a protocol.

### DIFF
--- a/pixie/walk.pxi
+++ b/pixie/walk.pxi
@@ -1,0 +1,113 @@
+(ns pixie.walk)
+
+(defprotocol IWalk
+  (-walk [x f]))
+
+(extend-protocol IWalk
+  PersistentList
+  (-walk [x f]
+    (apply list (map f x)))
+
+  Cons
+  (-walk [x f]
+    (cons (f (first x)) (map f (next x))))
+
+  IMapEntry
+  (-walk [x f]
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
I've restarted the build as it looks like pulling down deps failed.

mpenet:
yep, the pypy archive is not available https://bitbucket.org/pypy/pypy/get/default.tar.bz2 returns an http 500 at the moment.

mpenet:
I forgot to mention, this API doesn't include the "outer" fn parameter from the clojure equivalent, I never really understood why they baked in this fn, as you can just call it on the result returned by a walk yourself and seems to make things more confusing for little benefit. 

Feel free to tell me if you think it's not ok.

thomasmulvaney:
I agree that the outer is probably not necessary. I like the choice of an `IWalk` protocol so walking can be extended in the future.

mpenet:
Do you need me to modify anything on this?  (I need it for a toy project, but worst case I can use a local fork).

thomasmulvaney:
A quick squash would be nice, but otherwise this looks great. Thanks!

mpenet:
There you go! Squashed.

heyLu:
Thanks!

PR #429, opened 2015-11-03
Improve support for additional load-paths

emallson:
This PR was motivated by me wanting to create an [AUR](https://aur.archlinux.org) package for pixie, but being blocked by the hard-coded `load-paths`. 

This PR contributes the following changes:

- load-paths are now added from the `PIXIE_PATH` environment variable when `pixie-vm` is run
- A single load path is added during build. This is intended for use by packagers to allow the standard library to be installed to a system directory such as `/usr/share/pixie`
- A `make install` target that more-or-less conforms to the variable conventions.

The order of the load-paths is [user path (-l), pixie-vm directory, PIXIE_PATH, system path].

Note that the pixie-vm path seems weird when doing a system install, because `/usr/bin` and a bunch of directories related to it end up in `load-paths`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,13 +42,13 @@ compile_basics:
 	@echo -e "\n\n\n\nWARNING: Compiling core libs. If you want to modify one of these files delete the .pxic files first\n\n\n\n"
 	./pixie-vm -c pixie/uv.pxi -c pixie/io.pxi -c pixie/stacklets.pxi -c pixie/stdlib.pxi -c pixie/repl.pxi
 
-build_preload_with_jit: fetch_externals
+build_preload_with_jit: fetch_externals standard_path.py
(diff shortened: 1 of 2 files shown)


heyLu:
Sorry for the late response...

While we need to support installing pixie separate from its libraries, I think we should do it "cleaner".

I'd prefer switching between the current path local to `pixie-vm` and a "system" path at build time.  Either one or the other, but not both, because the current load path is confusing enough as it is.

Something like:

```python
if USE_GLOBAL_LOADPATH:
  system_path = ... # from LIB_PATH or similar
else:
  system_path = ... # local to `pixie-vm`
```

emallson:
Sorry for the also-late response :P

Understood. I'm going to close this PR and use that approach. My reasoning for doing it this (messy) way was to preserve as much existing behavior as possible.

I should have a followup PR up this weekend.

PR #430, opened 2015-11-04
Basic doc generator

hswick:
This is meant to implement minimal doc generation capabilities.

Currently generates .rst files which can be viewed with command restview (pip install restview)

Displays metadata fields :docs, :examples, :line-number, and :signature.

Use the script  via ./pixie-vm pixie/generate-docs.pxi namespace > docs/namespace.rst


thomasmulvaney:
I think this could be really nice to get in now that as someone has gone to the effort of setting up a website with docs: http://docs.pixielang.org/en/latest/namespaces.html What do you think @heyLu and @halgari. Obviously a rebase would be nice and this can probably go into master rather than doc-gen. Maybe under `scripts/generate-docs.pxi` rather than under `pixie/`? Thoughts?

thomasmulvaney:
Oh and the generated `*.rst` docs look brilliant by the way @hswick! :+1:

heyLu:
@thomasmulvaney sounds good!

@hswick thanks for this!  Could you rebase this onto master and submit a PR for that?  Please move the script to either `generate-docs.pxi`, or `scripts/generate-docs.pxi` as @thomasmulvaney mentioned.

hswick:
@thomasmulvaney I'm glad you liked the docs. I'd be willing to do more work, just wasn't sure what was wanted. 
Also, should I focus on the html or the rst? As I saw @halgari messing around with both.
I submitted a new pull request for the doc generator so I'll close this one now.

PR #434, opened 2015-11-06
Fix sorting branch

thomasmulvaney:
FIxes stackoverflow in JIT mode with a trampoline. We should still look for a C lib that could do this stuff or at least do a dirty mutating approach which is more JIT friendly. For now this is at least a simple solution.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -2974,3 +2974,36 @@ ex: (vary-meta x assoc :foo 42)"
    :added "0.1"}
   [x f & args]
   (with-meta x (apply f (meta x) args)))
+
+;;; Sorting
+
+(defn -merge-sort-step
+  [comp-fn merged-left merged-right res]
+  (cond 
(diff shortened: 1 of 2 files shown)


heyLu:
Can you make the helper functions (`trampoline` and the `-merge` fns) private?

You also mentioned in #361 that you had implemented a version using arrays, do you still have that or did it not work out?

thomasmulvaney:
I agree the helpers should be private. Trampoline might be the exception as
it is rather useful. My mutating version got over ambitious when I tried
adding some of the TimSort optimizations - it's rather broken at the moment.

I'll make those adjusts tomorrow evening.

Cheers
On Nov 7, 2015 9:42 AM, "Lucas Stadler" <notifications@github.com> wrote:
>
> Can you make the helper functions (trampoline and the -merge fns) private?
>
> You also mentioned in #361 that you had implemented a version using
arrays, do you still have that or did it not work out?

PR #439, opened 2015-11-11
add csp/timeout channel

mpenet:
It's the same as the core.async version. 

I did it using pixie.uv direcly, there might be a smarter way to do this using pixie.stacklets but I wasn't sure how.


heyLu:
Thanks!

PR #441, opened 2015-11-13
Improves the performance of range dramatically

thomasmulvaney:
We make Range implement `ISeq`, giving it `-first` and `-next` methods, returning start and a new Range
respectively. This avoids all the less JIT friendly machinery of lazy sequences.

This change makes doing (doseq [i (range 1000000)] i) about 1000x faster on my machine.

We also make `last` more efficient by making a special case for `IIndexed` collections.


heyLu:
Thanks a lot, a 1000x speedup is always welcome! :)

PR #453, opened 2015-11-27
line-reader accepts a non buffered stream

thomasmulvaney:
We create a buffered stream with a buffer size of one.


heyLu:
Thanks! :)

Issue #455, opened 2015-12-01
support nix package


bendlas:
I've packaged pixie with [nix](http://nixos.org/nix/): https://github.com/bendlas/nixpkgs/tree/a15e55bdbe956273f0be3915f8c325a958080fb2/pkgs/development/interpreters/pixie (the .patch file, in particular)

I think that nixos can offer tremendous value to pixie:
- binary distribution (nix can work within another distro, like npm)
- stable dependencies
- ease of configuring vms, docker images, ...

In order to get pixie built, I had to do some work, that I'd like to offer back upstream. This will also benefit other distributions, that would like to package pixie:
- I basically recreated the whole makefile as a nix expression. I had to do this, because the included makefile seems pretty ad-hoc, downloading unversioned dependecies, creating stuff in the parent directory. I'd love to see a more packaging-friendly makefile, so that I can reduce the complexity of the build expression. But more importantly, I had to patch pixie's source for two issues and we should focus on getting them patched first:
- pixie hard-codes pathes relative to its executable, as load-paths. Its load-paths are also the only place it looks for c header-files and library objects. My current patch replaces those relative paths with stable (crypto-hash) pathes to the c-dependencies it mentions in the docs (+ zlib, which it doesn't document). A proper solution should have a configuration phase, where it spits the pathes from pkg-config into some sort of `config.edn`
- pixie's ffi generates a file `/tmp/tmp.cpp` and compiles it. This is not concurrency-safe, as well as a security hole + it doesn't work when /tmp is not writable (as is the case within a nix-build). My patch replaces this with a usage of (getenv "TMPDIR") + mkdtemp to create a temporary directory to compile its ffi info in.

It should be clear, that all of this are issues, not just concerning nixos, but any distribution, really.
However, I didn't create a pull request right away, because I'd like to get feedback from pixie's maintainers, before reworking its build process. If you are interested and also willing to keep half an eye on integration, I'd be willing push some patches to you, as well as get pixie into the nixos distribution.

bendlas:
If you'd like to build if for youself, make sure you've installed the nix package manager, and then:
```
git clone https://github.com/bendlas/nixpkgs.git
cd nixpkgs
git checkout build-pixie
nix-build . -A pixie
```

heyLu:
Thanks for doing this!

Can you create a pull request for the ffi tmpfile patch?

The load path issue is more complicated...  I don't really like having to hard code these paths, I posted some thoughts on this in #429.

[...]

PR #458, opened 2015-12-03
Adding a few missing trig functions.

cgore:

PR #459, opened 2015-12-04
Improve tempdir usage in ffi-infer

bendlas:
- Make use of mkdtemp to create unique temporary directories per call to
  run-infer
- If $TMPDIR is set, use it instead of /tmp
- Remove temp files after finishing

see #455

### DIFF
--- a/pixie/ffi-infer.pxi
+++ b/pixie/ffi-infer.pxi
@@ -178,24 +178,35 @@ return 0;
   `(def ~(symbol name)
      ~(callback-type of-type false)))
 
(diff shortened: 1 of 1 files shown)


heyLu:
The failure is the same as in #458, because something on the pypy side changed.  (Not sure what, and I didn't have time yet to figure out what it might be.)

bendlas:
I suggest fixing the pypy version to the latest release: https://bitbucket.org/pypy/pypy/downloads/pypy-4.0.1-src.tar.bz2 That's what I did with the nix package.
We can still deal with the problem if it persists with the next release.

heyLu:
Thanks!

PR #460, opened 2015-12-04
Add CFloat ffi type.

nilern:
This is useful for interop with OpenGL (for example). 

As an aside, maybe the nomenclature should be `CFloat32` and `CFloat64` instead of `CFloat` and `CDouble`?


thomasmulvaney:
If we could get some basic tests around this I'd be happy to merge this.

nilern:
I Added some tests involving `math.h`. Didn't find any explicit tests for `CDouble` either.

thomasmulvaney:
Thanks!

PR #462, opened 2015-12-05
Add `is_recursive=True` to the main JitDriver

heyLu:
This seems to be required in recent versions of pypy/rpython.

PR #465, opened 2015-12-07
Revert "Add `is_recursive=True` to the main JitDriver"

heyLu:
Reverts pixie-lang/pixie#462 as this broke with *another* change in pypy which will be fixed in #464.


heyLu:
Closing as this won't fix the build either. #464 will.

PR #467, opened 2015-12-07
parser.json : add write-string for generating JSON

jmglov:


jmglov:
This is probably not the best way to do JSON writing, but I figured I'd open a PR to get some feedback and get the ball rolling. I'll happily implement suggestions. :)

thomasmulvaney:
What do you think about using a protocol? It seems you are dispatching based on types most of the time so extending types to implement a `IToJSON` protocol maybe better than multimethods? Looks good though!

jmglov:
I'll give the protocol approach a go and see how that turns out. I was also thinking that if there's a pure-Clojure JSON implementation out there, maybe I can make Pixie compile it.

PR #469, opened 2015-12-08
Add pixie.data.json for reading and writing JSON

jmglov:
@thomasmulvaney I like the way this turned out. Thanks for all of the helpful feedback!

### DIFF
--- a/pixie/data/json.pxi
+++ b/pixie/data/json.pxi
@@ -1,0 +1,40 @@
+(ns pixie.data.json
+  (:require [pixie.parser.json]
+            [pixie.string :as string]))
+
+(def read-string pixie.parser.json/read-string)
+
+(defprotocol IToJSON
+  (write-string [this]))
+
+(defn- write-map [m]
+  (if (empty? m)
+    "{}"
+    (str "{ "
+         (->> m
+              (map (fn [[k v]] (string/interp "$(write-string k)$: $(write-string v)$")))
+              (string/join ", "))
+         " }")))
+
+(defn- write-sequential [xs]
+  (if (empty? xs)
+    "[]"
+    (str "[ "
+         (->> (map write-string xs)
+              (string/join ", "))
+         " ]")))
+
+(defn- write-str [s]
+  (string/interp "\"$s$\""))
+
+(extend-protocol IToJSON
+  Character (write-string [this] (write-str this))
+  Cons (write-string [this] (write-sequential this))
+  EmptyList (write-string [this] (write-sequential this))
+  Number (write-string [this] (str this))
+  Keyword (write-string [this] (write-str (name this)))
+  LazySeq (write-string [this] (write-sequential this))
+  IMap (write-string [this] (write-map this))
+  IVector (write-string [this] (write-sequential this))
+  Ratio (write-string [this] (str (float this)))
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
Very nice! One improvement would be to use transduces in that map step. Something like: https://github.com/pixie-lang/pixie/blob/master/pixie/stdlib.pxi#L348 I'm guessing using a transducer should be a faster than just a lazy sequence. But that's perhaps something to do later.

One thing that *would* be nice is a test where you take some data, write it to json, read it, and make sure it's the same as data.

jmglov:
@thomasmulvaney Good comments, both. Let's see if I can address them both in 15 minutes before I go home. ;)

jmglov:
@thomasmulvaney OK, I've incorporated your suggestions.

thomasmulvaney:
Looks like you need to extend PersistentList as well.

jmglov:
@thomasmulvaney I'm sure you're right, but would you mind providing a line of code that demonstrates the problem for me? I'd like to add it as a test case. :)

thomasmulvaney:
The travis build failed with https://travis-ci.org/pixie-lang/pixie/jobs/95605505#L1605
When I get green light from travis i'll merge this.

Thanks for working so hard on it!

thomasmulvaney:
Ok sorry, i've just discovered how that error arises. When you compile a pxi file to pxic `Cons` lists become `PersistentList`s. As the travis build compiles all tests first so the `'(1 2 3)` ended up becoming `(list 1 2 3)`. I've never run into that before.

jmglov:
@thomasmulvaney Good catch! I should have looked at the Travis error myself. I just made the (faulty) assumption that it was the pypy build bug. Ugh.

Sorry about all the careless mistakes. As I get used to the Pixie workflow, I'll hopefully get better.

thomasmulvaney:
Not a problem, I wouldn't have picked up on that PersistentList bug either. I'm not even sure if its expected behaviour tbh. Perhaps we should add a make task for running tests both uncompiled and compiled to make local testing that little bit more thorough.

jmglov:
```
user => (instance? PersistentList (list 1 2))
true
user => (instance? PersistentList '(1 2))
false
```

I think we might want to think about changing this. To my old-school mind at least, there should be no difference between `(list ...)` and `'(...)`.

Perhaps @halgari can provide some insight into this; maybe it's by design and for a very good reason.

jmglov:
@thomasmulvaney I did the obvious yet previously overlooked thing and copied the `parser.json` tests into the `data.json` test file and reversed them. Good thing too, because I wasn't handling `nil`! Anyway, I'm pretty happy with the code now. Let's see if Travis agrees. ;)

thomasmulvaney:
@jmglov I would make an issue around that unexplained PersistentList list behaviour. If you look into the reader you'll see it creates an `EmptyList` when given `'()` which is what `(list)` also creates.

jmglov:
@thomasmulvaney Yes, `EmptyList` is what I would expect from both. I'll file an issue for the other.

This PR is green now. :)

thomasmulvaney:
Great! Thanks for working on this!

heyLu:
:tada:

PR #479, opened 2015-12-21
Makefile : do not fetch externals when present

jmglov:


thomasmulvaney:
Thanks! Been wishing we had this for ages!

Issue #483, opened 2016-01-07
Add an eduction type and function


thomasmulvaney:
Add an `Eduction` type and function to `stdlib.pxi`.

alekcz:
@thomasmulvaney what's an `Education` type?

thomasmulvaney:
@alekcz an `Eduction` is basically the object which is returned by calling `eduction`: http://clojure.github.io/clojure/clojure.core-api.html#clojure.core/eduction . It essentially wraps a `xform` function (just like what one can pass `into` or `transduce`) as well as a collection and provides a `IReducible` interface. It's sort of like a suspended collection + transformation. The Pixie version wouldn't have to support `Iterable` as we don't currently have that.

thomasmulvaney:
Added thanks to @sebastiano-barrera

PR #492, opened 2016-02-03
Add `eduction` and `completing`

sebastiano-barrera:
The `eduction` function was added, as requested in issue #483.

The implementation and the tests were imported and adapted from
Clojure's standard library (from [1] and [2], respectively).

Since Clojure's implementation uses it, the function `completing` was
imported as well (from the same source).

As of now, Pixie's stdlib lacks a `sort`/`sort-by` function, one of the
tests is disabled (commented out), and another is modified in order to
avoid the call to `sort-by`.

Sources:
Both from Clojure's main repo, fetched at commit `3b98a00`.
[1] https://github.com/clojure/clojure/blob/3b98a00e86f961550fb2eaee64e70754e04d1089/src/clj/clojure/core.clj#L7339
[2] https://github.com/clojure/clojure/blob/3b98a00e86f961550fb2eaee64e70754e04d1089/test/clojure/test_clojure/transducers.clj

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -1884,6 +1884,35 @@ not enough elements were present."
             (when-let [s (seq coll)]
               (reductions f (f init (first s)) (rest s))))))))
 
+(defn completing
+  "Takes a reducing function f of 2 args and returns a fn suitable for
+  transduce by adding an arity-1 signature that calls cf (default -
+  identity) on the result argument."
+  ([f] (completing f identity))
+  ([f cf]
+     (fn
+       ([] (f))
+       ([x] (cf x))
+       ([x y] (f x y)))))
+
+(deftype Eduction [xform coll]
+   IReduce
+   (-reduce [self f init]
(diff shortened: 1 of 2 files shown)


sebastiano-barrera:
I have a slight doubt: are the two test suites (Clojure's original test suite and the one that this PR adds, where `sort`s are avoided) really equivalent? My understanding is that transducers respect the order of the items in the input collection (making the calls to `sort` serve only an aesthetic purpose), but I'm in no way an expert, so I'd like someone else to comment on this.

thomasmulvaney:
Without sort the output of the example would be `[1 1 2 1 2 3 1 2 3 4]` which is the concatenation of the ranges 1 - 4: `(concat (range 1 2) (range 1 3) (range 1 4) (range 1 5))`.  If you change the test to check against the output printed above rather than sorted version that would be great. Otherwise looks good! Thanks!

sebastiano-barrera:
Oh I see it now... it also becomes exactly identical to the test right after it! I'll push the commit without that test right away

thomasmulvaney:
Thanks!

PR #493, opened 2016-02-09
make typing less verbose

thomasmulvaney:
Object's type method is now used by default so all
subtypes can use the inherited implementation


heyLu:
Looks good, thanks! :)

PR #495, opened 2016-02-12
Added iterate function and tests

oholiab:
A couple of things about this PR:

* this function is in clojure stdlib so hopefully is suitable for pixie stdlib
* I wasn't sure what the convention was on placement so I just stuck it on the bottom (same with the tests)
* I lifted the docstring wholesale from the clojure one - I don't know if that's a good thing or a bad thing, I'm happy to reword it if that causes an issue.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -3072,3 +3072,11 @@ ex: (vary-meta x assoc :foo 42)"
             (swap! cache assoc argsv ret)
             ret)
           val)))))
+
+(defn iterate
+  {:doc "Returns a lazy sequence of x, (f x), (f (f x)) etc. f must be free of
+    side-effects"
+   :signatures [[f x]]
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
Nice one. I would suggest that `iterate` should implement `IReduce` as well though. I'd have a glance at how `repeat` or `range` are defined for inspiration.

oholiab:
Cool, I'll get on it :)

thomasmulvaney:
This is a good `LazySeq` version of `iterate`. If you're happy for it to be merged as is, I'll add an enhancement issue noting that adding `IReduce` support would be useful.

oholiab:
I've refactored it into a type and I've done a really basic IReduce for it... How does that look?

thomasmulvaney:
Looks good to me! Thanks for working on this!

PR #496, opened 2016-02-13
find python using which -a

keymone:
python2 does not exist on osx, use `env which -a python2 python2.7` if none provided


thomasmulvaney:
Thanks!

PR #499, opened 2016-02-18
Added mkdtmp to stdlib

oholiab:
This is just a copy of the same def in ffi-infer, just seems like it should be in stdlib (and I want to use it ;) )


oholiab:
Also just added rmdir as one is not so helpful without the other

heyLu:
Thanks!  Can you add `mkdir` as well?  :)

oholiab:
Sure :)

oholiab:
I added rm whilst I was at it seeing as I need that too ;)

thomasmulvaney:
Merged :sparkles:

PR #500, opened 2016-02-19
fixes the speed of iterates reduce

thomasmulvaney:
@oholiab the reduce implementation was creating some lazyseqs which was making it not as performant as it should have been.


oholiab:
Nice :) Thanks for keeping me in the loop!

PR #508, opened 2016-03-07
fix tail call check for recur

### LINKED ISSUE #440: Recur not throwing exception
The following code should throw an exception because recur is not called from a tail position. Instead it always returns 1.

```Clojure
(defn factorial [n]
  (if (zero? n)
    1
    (* n (recur (dec n)))))
```

thomasmulvaney:
This fixes #440

### DIFF
--- a/pixie/vm/compiler.py
+++ b/pixie/vm/compiler.py
@@ -539,8 +539,8 @@ def compile_fn_body(name, args, body, ctx):
         compile_form(body, new_ctx)
     else:
         while body is not nil:
-            if rt.next(body) is nil:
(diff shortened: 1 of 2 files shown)


thomasmulvaney:
I've removed a bunch of unused operations here as well. The real important changes are in compile_fn_body, and of course enabling the assertion that recur is in tail position.

thomasmulvaney:
Also, we return a `RuntimeException` but maybe we should introduce a `CompilerException`?

thomasmulvaney:
I've removed the extra commits which deleted dead code for now to keep this PR straight forward.

Issue #502, opened 2016-02-27
Implement sequence


thomasmulvaney:
It would be nice to have a `sequence` function which accepts either a transducer and a collection or an eduction and returns a lazy sequence.

thomasmulvaney:
Closed in #509. Thanks @angusiguess

PR #509, opened 2016-03-09
Add sequence to stdlib

angusiguess:
Implement sequence as per #502.


thomasmulvaney:
Thanks for working on this. Would you be able to add some tests please?

angusiguess:
Sure thing. While I was testing I noticed that eductions break the seqable contract in the empty case.

We'd expect:
```(seq (eduction (map inc) '[]))``` to produce nil

but it produces ```[]```

Calling seq on the result of the reduce seems to resolve this.

thomasmulvaney:
I think the problem is that the `-seq` implementation on `Eduction` just does a reduce rather than returning a `LazySeq`. I think the best thing to do would be to fix that by making it lazy.

angusiguess:
Agreed, I'll take a crack at it.

angusiguess:
So I inverted things a little bit by making ```sequence``` work lazily and eduction make use of ```sequence```

thomasmulvaney:
Thanks for working on this!

PR #531, opened 2017-06-08
Change calls to throw to vector instead of string

egregius313:
Calls to throw previously used strings, which is not a valid way to call throw (Thus raising an exception raised a completely different exception). Changed all calls to use vectors instead.

### DIFF
--- a/pixie/fs.pxi
+++ b/pixie/fs.pxi
@@ -109,7 +109,7 @@
   (rel [this other]
     (if (satisfies? IFSPath other)
       (rel-path this other)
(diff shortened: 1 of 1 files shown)


heyLu:
This looks good, but I'm not sure about the custom exceptions.  What do other languages do here?

thomasmulvaney:
Java has some file exceptions which subclass [IOException](https://docs.oracle.com/javase/7/docs/api/java/io/IOException.html). It certainly is nice to be handle exceptions on a case by case basis.

heyLu:
Thanks @egregius313! :sparkles:

PR #528, opened 2017-03-11
Added letfn macro to the stdlib.

egregius313:
I find I regularly use the letfn macro and think the most important place for it is the stdlib.

PR #534, opened 2017-10-05
Add test for the letfn macro

egregius313:
Sorry, I realized I never actually added the tests for the `letfn` macro, and for some reason I can't figure out how to add the test to the test file in #528.


heyLu:
Thanks! :)

Issue #535, opened 2018-01-08
RuntimeException: <inst pixie.stdlib.Keyword> Var assoc is undefined


rrrnld:
I just tried compiling pixie on a Macbook Pro, macOS 10.13.2, but couldn't run the resulting `pixie-vm` executable. It crashed with the following error message: 
```
$ ./pixie-vm 
in internal function run_load_stdlib

in internal function load-ns

in internal function load-file

in internal function load-reader

RuntimeException: <inst pixie.stdlib.Keyword> Var assoc is undefined
```

Following the instructions to build the executable all I basically did was:

```
$ brew install libuv libffi boost
$ git clone https://github.com/pixie-lang/pixie.git
$ make build_with_jit
```

Here's some additional info about the installed versions:

```
$ brew info libuv libffi boost
libuv: stable 1.18.0 (bottled), HEAD
Multi-platform support library with a focus on asynchronous I/O
https://github.com/libuv/libuv
/usr/local/Cellar/libuv/1.18.0 (56 files, 2.8MB) *
  Poured from bottle on 2018-01-08 at 10:12:42
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libuv.rb
==> Dependencies
Build: pkg-config ✔, automake ✘, autoconf ✔, libtool ✔, sphinx-doc ✘
==> Options
--with-test
	Execute compile time checks (Requires Internet connection)
--HEAD
	Install HEAD version

libffi: stable 3.2.1 (bottled), HEAD [keg-only]
Portable Foreign Function Interface library
https://sourceware.org/libffi/
/usr/local/Cellar/libffi/3.2.1 (16 files, 296.9KB)
  Poured from bottle on 2017-07-01 at 22:57:15
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libffi.rb
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local,
because some formulae require a newer version of libffi.

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/libffi/lib
For pkg-config to find this software you may need to set:
    PKG_CONFIG_PATH: /usr/local/opt/libffi/lib/pkgconfig


boost: stable 1.66.0 (bottled), HEAD
Collection of portable C++ source libraries
https://www.boost.org/
/usr/local/Cellar/boost/1.66.0 (13,101 files, 435.2MB) *
  Poured from bottle on 2018-01-08 at 10:12:59
[...]

PR #538, opened 2019-01-02
Revert back to a working version

kuettler:
The latest addition to stdlib seems to break pixie startup. Going back allows a clean build to succeed so pixie can be started.

### DIFF
--- a/pixie/stdlib.pxi
+++ b/pixie/stdlib.pxi
@@ -482,19 +482,6 @@
   (let [nm (with-meta nm (assoc (meta nm) :private true))]
     (cons `defn (cons nm rest))))
 
-(defmacro letfn
-  "fnspec ==> (fname [params*] exprs) or (fname ([params*] exprs)+)
-
-  Takes a vector of function specs and a body, and generates a set of
-  bindings of functions to their names. All of the names are available
-  in all of the definitions of the functions, as well as the body."
-  {:added "0.2" 
-   :forms '[(letfn [fnspecs*] exprs*)]}
-  [fnspecs & body]
-  `(letfn* ~(vec (interleave (map first fnspecs)
-                             (map #(cons `fn %) fnspecs)))
-           ~@body))
-               
 (defn not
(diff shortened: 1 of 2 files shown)


oubiwann:
For context on this PR, see https://github.com/pixie-lang/pixie/issues/535#issuecomment-357545899

thomasmulvaney:
Ok it's been awhile! 

When I was playing with Pixie alot it was when I wasn't working full time. Unfortunately I still don't have much time spare.

There are other people with commit rights to this repository such as the creator @halgari  and @heyLu  but I'm not sure how much time they have either. I got a lot out of playing round with it, hopefully others will too! I guess that means giving those who have time commit rights or maybe just maintaining your own forks.

In regards to getting this merged:

The PR that broke things was #534 which passed the CI. Is there something which we are not testing for currently, that should be added to the test suite that would have caught that issue? If so, can it be added to the tests?

oubiwann:
I’d love to help get things in working order for this repo again — no only the revert, but testing a branch with the broken change to make sure the build targets we’re using in Travis can catch issues like that in the future.

I can even take a look at `let*` to see about getting `letfn` back in ...

I think it would be nice for the project if this work could happen here ... it would help give interested Lisp explorers a better experience with Pixie :-)

alekcz:
@thomasmulvaney the CI build doesn't fail when a runtime exception is thrown. That's probably how the error snuck in. See https://github.com/pixie-lang/pixie/issues/535#issuecomment-357068079 .

I'm super keen to get involved. I think at one point I even mail @halgari

thomasmulvaney:
@oubiwann Given I don't even have Pixie environment to play with atm, I'm relying on the CI doing its job. So yes, lets get that in shape.

@alekcz mentioned the tests don't fail when there is a runtime exception. Would someone like to open a PR which just makes the current tests fail when this happens? Then this PR passing will actually mean something.

g3ntleman:
Do we have to wait for the CI-Test to get improved before we get master building again?
master not building could scare-off new people (like me).

PR #540, opened 2020-10-28
Treat .pxi files as vendored in language statistics

pchaigno:
GitHub's language detection library, Linguist, [associates the `.pxi` file extension with Cython](https://github.com/github/linguist/blob/143cd81b3e1a70f094f094a34498747211fecfc3/lib/linguist/languages.yml#L1074). This results in incorrect language statistics and syntax highlighting for `.pxi` files on the repository. This commit fixes it by using [Linguist overrides](https://github.com/github/linguist/#vendored-code) to tell Linguist to ignore those files.

Another option would be to recognize those files as some Lisp variant. Linguist has a few options (Common Lisp, PicoLisp, NewLisp, and Emacs Lisp). I tested each of them using [Lightshow](https://github-lightshow.herokuapp.com/), but the syntax highlighting is never great.


heyLu:
Hi!  Thanks for the links to the linguist docs, that helps a lot!

I think it would be better to recognize the language as Clojure, because that's the original inspiration for the language, and probably the closest one.

From the [linguist docs](https://github.com/github/linguist/#using-gitattributes) I think something like the following would be better:

```
*.pxi linguist-language=Clojure
```

@pchaigno, out of interest, are you using Pixie yourself?  It's pretty much unmaintained since a few years, but it'd be neat if you are using it. :)

